PR #215, opened 2026-04-13
feat: make ingress/storage config cloud-agnostic and add provider CI profiles

### LINKED ISSUE #180: [Issue]: Helm Chart - Abstract Storage and Ingress Configuration
## Summary
Make Helm chart storage and ingress settings cloud-agnostic by removing hardcoded provider-specific classes/annotations and exposing them via `values.yaml`.

## 🧩 Type of Issue
✨ Feature request

## 🔍 Description
As a DevOps engineer, I want to configure custom `StorageClass` and Ingress class settings in the Helm release so the application can be deployed on any Kubernetes cluster regardless of cloud provider.

Current Helm templates may contain cloud-provider-specific classes or annotations. We should audit and refactor PVC and Ingress templates so these settings are fully configurable via `values.yaml`.

Acceptance criteria:
- [ ] All PersistentVolumeClaims (PVCs) use a variable for `storageClassName` (defaulting to `null` or cluster default).
- [ ] Ingress templates use a variable for `ingressClassName`.
- [ ] Ingress annotations are fully overrideable and mergeable from `values.yaml`.
- [ ] Default values are compatible with standard Kubernetes distributions (Minikube, Kind) and managed cloud clusters.
- [ ] CI pipeline passes linting with different cloud-provider configurations.

## 🧪 Steps to Reproduce (for bugs)
N/A (feature request)

## 💻 Environment
- Kubara version: current main branch
- Kubernetes version: multiple targets (Kind, Minikube, managed cloud)
- Helm version: current supported version

## 📎 Additional Context / Logs
Potential implementation detail:
- Add provider-specific values files for CI matrix linting (for example `values-kind.yaml`, `values-eks.yaml`, `values-gke.yaml`).

Matthiator:
## 📝 Summary

Closes #180

This PR improves ingress/storage configurability for cloud-agnostic deployments while keeping kubara's default behavior stable.

What changed:
- Restored `ingressClassName` default behavior (`traefik`) for compatibility and out-of-the-box UX.
- Templates still render ingress class conditionally, and users can override `ingressClassName` for other controllers.
- Added `PersistentService` with per-service `storageClassName` (used by `kubePrometheusStack` and `loki`).
- `storageClassName` is rendered only when set; otherwise cluster/default chart behavior is used.
- Added per-service ingress annotation overrides via `services.<service>.ingress.annotations`.
- Default annotations are merged with user annotations (user values win on key conflicts).
- Traefik-specific middleware annotations are rendered only when both `oauth2Proxy` and `traefik` are enabled.
- The Kyverno Policy Reporter Traefik Middleware CRD is guarded by the same condition to avoid CRD/sync issues on non-Traefik clusters.

Implementation refinement after review:
- Removed the temporary cmd-side computed ingress helper (`generate_context.go`) and moved annotation composition back to template-local logic.
- Kept PR scope focused; larger architecture refactors are tracked as follow-up issues.

CI profiles:
- Added provider overlays (`kind`, `eks`, `gke`) for ingress-enabled services (`argo-cd`, `homer-dashboard`, `oauth2-proxy`, `longhorn`, `kube-prometheus-stack`, `loki`, `kyverno-policy-reporter`).
- CI overlays are now stored under `.github/helm-profiles/<chart>/values-*.yaml`.
- They are used by the PR workflow and are not generated into customer output.

## 🧩 Type of change

- 🔧 CLI / Go code
- 📦 Helm chart / templating
- 📝 Documentation
- 🧪 Test or CI change
- ♻️ Refactor / cleanup

- No

Behavior note:
- Default ingress class remains `traefik` (compatible with existing behavior).
- Non-Traefik environments can explicitly set `ingressClassName` and controller-specific ingress annotations.

## 🧪 Testing

- Unit tests passed locally:
  - `cd go-binary && go test ./...`
- Static checks passed locally:
  - `cd go-binary && go vet ./...`
  - `cd go-binary && golangci-lint run` (via `make lint`)
- CI pipeline checks passed for this PR.

## 🔗 Related Issues / Tickets

Closes #180

Follow-up issues created from review:
- #217 CI: Extract PR validation bash blocks into reusable scripts
- #218 Refactor generate command logic into internal/generate package
- #219 Add service lifecycle/registration extensibility for templating

## 📎 Additional Context

Annotation merging is additive:
- user-provided annotations overwrite same-key defaults
- defaults with keys not overridden remain in place
- this behavior is documented in `bootstrap_process.md` and `add_worker_cluster.md`


Matthiator:
Partially related to #182.

This PR updates rendering logic so Traefik-specific auth annotations/CRDs are only generated when both `oauth2Proxy` and `traefik` are enabled.
That improves behavior when oauth2-proxy is disabled, but does not fully close #182.

Remaining scope for #182:
- explicit test case for `oauth2Proxy: disabled`
- dedicated documentation of side effects and required follow-up changes when disabling oauth2-proxy

.github/workflows/pr-checks.yaml
```diff
@@ -323,6 +323,12 @@ jobs:
       run:
         shell: bash
     steps:
+      - name: Checkout CI overlay profiles
+        uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            go-binary/templates/ci-overlays
```
tuunit: I think the ci-overlays should live inside the `.github/**` directory and not inside the code as this will at some of point bleed into the binary

.github/workflows/pr-checks.yaml
tuunit, line 374: All this logic keeps growing and cannot be verified without running the full CI. Please create a new issue to refactor the enormous amounts of bash scripts into files that can be parameterised and tested / used without always running the full CI

.github/workflows/pr-checks.yaml
```diff
@@ -412,6 +418,25 @@ jobs:
               fi
               echo "📄 Rendering chart with overlay values"
               helm template --kube-version "$KUBE_VERSION" "${HELM_TEMPLATE_API_ARGS[@]}" --include-crds "$CHART" "$MANAGED_CHART_PATH" -f "$CUSTOMER_VALUES_FILE" > "$OUT_DIR/render.yaml"
+
+              PROFILE_DIR="${{ env.CODE_PATH }}/templates/ci-overlays/helm/$CHART"
+              if [[ -d "$PROFILE_DIR" ]]; then
+                shopt -s nullglob
+                PROFILE_FILES=("$PROFILE_DIR"/*.yaml)
+                shopt -u nullglob
+                for PROFILE_FILE in "${PROFILE_FILES[@]}"; do
```
tuunit:
```suggestion
                for PROFILE_FILE in "$PROFILE_DIR"/*.yaml; do
```

this should work

.github/workflows/pr-checks.yaml
```diff
@@ -421,20 +446,26 @@ jobs:
...
-                -schema-location https://json.schemastore.org/{{.ResourceKind}}.json  \
-                "$OUT_DIR/render.yaml" \
-                | tee "${{ env.REPORT_DIR }}/helm/${CHART}-kubeconform.log"; then
-              echo "::error file=$OUT_DIR/render.yaml::kubeconform failed for $CHART"
-              FAILED_CHARTS+=("$CHART")
-            fi
+            shopt -s nullglob
+            RENDER_FILES=("$OUT_DIR"/render*.yaml)
+            shopt -u nullglob
+            for RENDER_FILE in "${RENDER_FILES[@]}"; do
```
tuunit:
```suggestion
            for RENDER_FILE in "$OUT_DIR"/render*.yaml; do
```

go-binary/assets/config/configMngr_test.go
```diff
@@ -56,20 +56,20 @@ func newValidTestConfig() *Config {
...
+					Traefik:             GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					Kyverno:             GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					KyvernoPolicies:     GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					KyvernoPolicyReport: GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					Loki:                PersistentService{GenericService: GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}}},
+					HomerDashboard:      GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					Oauth2Proxy:         GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					MetricsServer:       GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					MetalLb:             GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
+					Longhorn:            GenericService{ServiceStatus: ServiceStatus{Status: StatusEnabled}},
```
tuunit, line 72: Why this change? As far as I can tell ServiceStatus is still an inlined element of GenericService

go-binary/assets/config/configMngr_test.go
```diff
@@ -532,7 +532,7 @@ clusters:
 	c := cm.GetConfig().Clusters[0]
 	assert.Equal(t, "dev", c.Stage, "Stage should be defaulted")
 	assert.Equal(t, "controlplane", c.Type, "Type should be defaulted")
-	assert.Equal(t, "traefik", c.IngressClassName, "IngressClassName should be defaulted")
+	assert.Empty(t, c.IngressClassName, "IngressClassName should remain empty when not set")
```
tuunit: Do we want that? Don't we expect the default to be set to traefik. At least that is what I implemented as part of #167

go-binary/assets/config/configTypes.go
```diff
@@ -15,7 +15,7 @@ type Cluster struct {
 	SSOOrg  string `json:"ssoOrg,omitempty" yaml:"ssoOrg,omitempty" jsonschema:"title=SSO Organization,description=The SSO organization or group allowed to access this cluster.,minLength=1"`
 	SSOTeam string `json:"ssoTeam,omitempty" yaml:"ssoTeam,omitempty" jsonschema:"title=SSO Team,description=The specific SSO team or sub-group allowed to access this cluster.,minLength=1"`
 
-	IngressClassName string `json:"ingressClassName,omitempty" yaml:"ingressClassName,omitempty" jsonschema:"title=Ingress Class,description=The ingress class to use for this cluster.,minLength=1,default=traefik"`
+	IngressClassName string `json:"ingressClassName,omitempty" yaml:"ingressClassName,omitempty" jsonschema:"title=Ingress Class,description=Optional ingress class to use for this cluster.,minLength=1"`
```
tuunit: Why remove the default? It is already optional because of the removal of `required` in #167 and will be set to `traefik` if empty

go-binary/assets/config/configTypes.go
```diff
@@ -67,8 +67,18 @@ type ServiceStatus struct {
 	Status Status `json:"status" yaml:"status" jsonschema:"title=Service Status,description=The desired status of the service.,enum=enabled,enum=disabled,default=disabled"`
 }
 
+type IngressOverrides struct {
+	Annotations map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty" jsonschema:"title=Ingress Annotations,description=Optional ingress annotations that are merged with kubara defaults."`
+}
+
 type GenericService struct {
 	ServiceStatus `yaml:",inline" jsonschema:"required,title=ServiceStatus"`
+	Ingress       *IngressOverrides `json:"ingress,omitempty" yaml:"ingress,omitempty" jsonschema:"title=Ingress Overrides,description=Optional ingress settings for this service."`
```
tuunit, line 76: Why make this a pointer? That will cause unnecessary nil checks down the line to avoid panics

go-binary/cmd/generate_context.go
tuunit, line 1: First of all, we shouldn't have such kind of logic inside the `cmd/` package. This package should only be concerned with user interaction and act as the user facing API. At the moment this package has way too many responsibilities. I think we should take the opportunity and start moving some of it into a separate package. Especially as this logic here is mutating the state of the processed it is best to move it out. 

My recommendation would be to have one package per command inside the internal package meaning we should start doing that for the generate cmd.

```
main.go
cmd/
  ...
internal/
  generate/  move business logic from cmd/generate.go and generate_context.go here
```

and we should get rid of all other packages at the root level but that will be a separate task and PR!

go-binary/cmd/generate_context.go
```diff
@@ -0,0 +1,99 @@
...
+
+	kyvernoPolicyReporterDefaults := baseIngressAnnotations(cluster, metalLbEnabled)
+	if traefikAuthEnabled {
+		kyvernoPolicyReporterDefaults[annotationTraefikRouterMiddleware] = traefikMiddlewareKyvernoPath
+		kyvernoPolicyReporterDefaults[annotationTraefikAppRoot] = traefikKyvernoAppRoot
+	}
+
+	longhornDefaults := standardIngressAnnotations(cluster, metalLbEnabled, traefikAuthEnabled)
+
+	oauth2ProxyDefaults := baseIngressAnnotations(cluster, metalLbEnabled)
```
tuunit: This is a lot of hard coding happening here. We need to generalize how we process services. This will and cannot scale! New services will always need to be manually added here and customer / users of kubara cannot easily add their own services through the kubara templating engine without building their own version from code.

Lets talk about this as part of the proposal discussion tomorrow. I think we could already benefit here by refactoring how the services are defined in `configTypes.go`. By changing the Services struct to map like so: `type Services map[string]Service` with a new `Service` type as a composition of the different services types we already have.

This way we could already build a loop over the services:

```golang
	for name, svc := range cluster.Services {
		if svc.Ingress == nil {
			continue
		}

		overrides := ingressOverrides(svc.Ingress)
		...
	}
```

go-binary/cmd/generate_context.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	merged := make(map[string]string, len(overrides.Annotations))
+	for key, value := range overrides.Annotations {
+		merged[key] = value
+	}
+	return merged
+}
+
+func mergeStringMaps(defaults map[string]string, overrides map[string]string) map[string]string {
+	merged := make(map[string]string, len(defaults)+len(overrides))
```
tuunit: this pre-allocation is a good idea but unfortunately not useful as you don't know how large the overlap is. worst case you allocate double the amount but overrides only has the elements that are inside defaults

go-binary/cmd/generate.go
```diff
@@ -164,8 +164,9 @@ func buildTemplateContext(clusterBlock config.Cluster, em envmap.EnvMap) (map[st
 	}
 
 	return map[string]any{
-		"cluster": clusterMap,
-		"env":     em,
+		"cluster":  clusterMap,
+		"env":      em,
+		"computed": buildTemplateComputed(clusterBlock),
```
tuunit: The naming scheme here seems wrong to me. The method only returns a single top level field `ingressAnnotations` so wouldn't it be better to do:


```suggestion
		"ingressAnnotations": generateIngressAnnotations(clusterBlock),
```

and remove the `ingressAnnotations` top level field from the build method?

go-binary/cmd/generate_context.go
```diff
@@ -0,0 +1,99 @@
...
+
+func baseIngressAnnotations(cluster config.Cluster, includeExternalDNSTarget bool) map[string]string {
+	annotations := map[string]string{
+		annotationCertManagerClusterIssuer: cluster.Services.CertManager.ClusterIssuer.Name,
+	}
+	if includeExternalDNSTarget {
+		annotations[annotationExternalDNSTarget] = cluster.PublicLoadBalancerIP
+	}
+	return annotations
+}
```
tuunit: fine for now but I think we need to introduce some kind of lifecycle hook system into kubara. so that services when they are loaded / registered for the templating they can "inject" into the the templating engine

go-binary/cmd/generate_context_test.go
```diff
@@ -0,0 +1,189 @@
...
+				GenericService: config.GenericService{
+					ServiceStatus: config.ServiceStatus{Status: config.StatusEnabled},
+				},
+			},
+			KyvernoPolicyReport: config.GenericService{
+				ServiceStatus: config.ServiceStatus{Status: config.StatusEnabled},
+			},
+			Longhorn: config.GenericService{
+				ServiceStatus: config.ServiceStatus{Status: config.StatusEnabled},
+			},
```
tuunit: Best to define a Services stub at the top of the file and reuse it in the test methods and only change the values that are relevant for the test. Or introduce a test builder pattern if you so desire

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
```diff
@@ -129,33 +129,31 @@ externalSecrets:
...
+
 argo-cd:
   server:
     ingressGrpc:
       enabled: true
       path: "/argocd"
-      ingressClassName: {{ .cluster.ingressClassName | default "traefik" }}
+{{- with .cluster.ingressClassName }}
+      ingressClassName: {{ . }}
+{{- end }}
```
tuunit:
```suggestion
      {{- with .cluster.ingressClassName }}
      ingressClassName: {{ . }}
      {{- end }}
```

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
```diff
@@ -129,33 +129,31 @@ externalSecrets:
...
       tls: true
       annotations:
-        cert-manager.io/cluster-issuer: {{ .cluster.services.certManager.clusterIssuer.name }}
-        {{- if (eq .cluster.services.metalLb.status "enabled") }}
-        external-dns.alpha.kubernetes.io/target: "{{ .cluster.publicLoadbalancerIP }}"
-        {{- end }}
+{{ toYaml $argocdIngressGrpcAnnotations | nindent 8 }}
```
tuunit:
```suggestion
        {{ toYaml $argocdIngressGrpcAnnotations | nindent 8 }}
```

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
```diff
@@ -129,33 +129,31 @@ externalSecrets:
...
     ingress:
       {{- if (eq .cluster.services.oauth2Proxy.status "enabled") }}
       enabled: true
       {{- end }}
       path: "/argocd"
-      ingressClassName: {{ .cluster.ingressClassName | default "traefik" }}
+{{- with .cluster.ingressClassName }}
+      ingressClassName: {{ . }}
+{{- end }}
```
tuunit:
```suggestion
      {{- with .cluster.ingressClassName }}
      ingressClassName: {{ . }}
      {{- end }}
```

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
```diff
@@ -129,33 +129,31 @@ externalSecrets:
...
       tls: true
       annotations:
-        cert-manager.io/cluster-issuer: {{ .cluster.services.certManager.clusterIssuer.name }}
-        {{- if (eq .cluster.services.oauth2Proxy.status "enabled") }}
-        traefik.ingress.kubernetes.io/router.middlewares: oauth2-proxy-oauth-auth@kubernetescrd
-        {{- end }}
-        {{- if (eq .cluster.services.metalLb.status "enabled") }}
-        external-dns.alpha.kubernetes.io/target: "{{ .cluster.publicLoadbalancerIP }}"
-        {{- end }}
+{{ toYaml $argocdIngressAnnotations | nindent 8 }}
```
tuunit:
```suggestion
        {{ toYaml $argocdIngressAnnotations | nindent 8 }}
```

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
tuunit, line 1: for readability purposes i personally would always indent the go templates to the same level as the rest of the yaml. otherwise it is hard to see where you are at. at least in my opinion

go-binary/templates/templates.go
```diff
@@ -43,6 +44,18 @@ var templateName = map[TemplateType]string{
...
+	funcs := sprig.FuncMap()
+	funcs["toYaml"] = func(v any) (string, error) {
+		out, err := yaml.Marshal(v)
+		if err != nil {
+			return "", err
+		}
+		return strings.TrimSuffix(string(out), "\n"), nil
+	}
+	return funcs
+}
```
tuunit, line 57: this feels off... what exactly is this for? why would we want to invent our own toYaml?

go-binary/cmd/generate_context.go
```diff
@@ -0,0 +1,99 @@
...
+
+	longhornDefaults := standardIngressAnnotations(cluster, metalLbEnabled, traefikAuthEnabled)
+
+	oauth2ProxyDefaults := baseIngressAnnotations(cluster, metalLbEnabled)
+
+	return map[string]any{
+		"ingressAnnotations": map[string]any{
+			"argocd": map[string]any{
+				"ingress": mergeStringMaps(argocdIngressDefaults, argocdOverrides),
+				"grpc":    mergeStringMaps(argocdGrpcDefaults, argocdOverrides),
```
tuunit: this is kind of ugly but i don't have better solution from the top of my mind. 🤔

tuunit (changes requested):

Matthiator:
Follow-up issues created for the larger architecture topics raised in review:

- #217 CI: Extract PR validation bash blocks into reusable scripts
- #218 Refactor generate command logic into internal/generate package
- #219 Add service lifecycle/registration extensibility for templating

I will keep this PR scoped to the concrete behavior/template fixes and track the broader refactors in these dedicated issues.

.github/workflows/pr-checks.yaml
Matthiator: Addressed in a005ced: CI overlay profiles were moved from `go-binary/templates/ci-overlays` to `.github/helm-profiles`, and the workflow path was updated.

.github/workflows/pr-checks.yaml
Matthiator: Agreed. I created follow-up issue #217 to extract the larger inline CI bash blocks into parameterized scripts that can be validated locally.

.github/workflows/pr-checks.yaml
Matthiator: Addressed in a005ced. The profile loop now uses `for PROFILE_FILE in "$PROFILE_DIR"/*.yaml; do` (with a safe existence guard).

.github/workflows/pr-checks.yaml
Matthiator: Addressed in a005ced. The render loop now uses `for RENDER_FILE in "$OUT_DIR"/render*.yaml; do` (with a safe existence guard).

go-binary/assets/config/configMngr_test.go
Matthiator: Good catch. This was partially mechanical fallout from introducing the new `Ingress` field on `GenericService`. Positional initialization no longer works cleanly here, so tests now use explicit field names.

go-binary/assets/config/configMngr_test.go
Matthiator: Addressed in a005ced: default behavior was restored. `ingressClassName` defaults to `traefik` again and tests were aligned.

go-binary/assets/config/configTypes.go
Matthiator: Addressed in a005ced: restored `default=traefik` on `Cluster.IngressClassName` to align with #167.

go-binary/assets/config/configTypes.go
Matthiator: I intentionally kept this as a pointer. In our generation flow (struct -> JSON -> map), a value type here tends to produce noisy `ingress: {}` output when not configured. Pointer keeps absent ingress overrides truly absent while templates safely access via `dig`.

go-binary/cmd/generate_context.go
Matthiator: Addressed for this PR scope: the newly added `generate_context.go` helper layer was removed. I opened follow-up #218 to track the broader `cmd` -> `internal/generate` refactor.

go-binary/cmd/generate_context.go
Matthiator: Addressed for this PR scope by removing the hardcoded service computation helper from Go and moving this back to template-local logic. Broader service extensibility is tracked in #219.

go-binary/cmd/generate_context.go
Matthiator: Resolved by removal of `generate_context.go` in a005ced; this pre-allocation logic no longer exists.

go-binary/cmd/generate.go
Matthiator: Resolved by removal of the `computed` wrapper/context in a005ced.

go-binary/cmd/generate_context.go
Matthiator: Agreed on direction. I opened #219 to track lifecycle/registration extensibility for services.

go-binary/cmd/generate_context_test.go
Matthiator: Resolved by removing `generate_context_test.go` together with that helper layer.

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
Matthiator: Addressed in a005ced. Template formatting was aligned and the conditional ingress class block is now in the suggested layout.

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
Matthiator: Addressed in a005ced. Formatting was aligned with the suggested indentation around `toYaml` output.

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
Matthiator: Addressed in a005ced. Formatting was aligned with the suggested indentation for the second ingress block as well.

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
Matthiator: Addressed in a005ced. Formatting was aligned with the suggested indentation around annotation rendering.

go-binary/templates/embedded/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
Matthiator: Agreed. Template block indentation/readability in this file was adjusted in a005ced.

go-binary/templates/templates.go
Matthiator: Good point. We keep this helper because kubara renders these files with Go `text/template` + sprig (not Helm runtime functions), and `toYaml` is required for map-based annotation rendering (`dict`/`set`/`mergeOverwrite`) used by this PR.

go-binary/cmd/generate_context.go
Matthiator: Addressed by removing the helper layer from `cmd` in a005ced and keeping the logic template-local for now.

Matthiator:
Design note on `ingressClassName` default

I want to clarify the decision in this PR from my side.

When I started implementing #180, my initial thought was to avoid a hardcoded ingress class default and allow omission of `ingressClassName` to make rendering more controller-agnostic.

After review and reconsidering the current kubara behavior, I decided to restore `default=traefik`.

My reasoning:
- kubara ships Traefik as the default ingress controller in the standard stack, so `traefik` as default is consistent with the out-of-the-box setup.
- this keeps behavior stable for existing users and upgrades (including expectations introduced in #167).
- flexibility is still preserved: users can explicitly set `ingressClassName` to other controllers (e.g. `nginx`, `alb`, `gce`) and disable Traefik.

So my final decision here is:
- keep `traefik` as the default,
- keep explicit override support for non-Traefik environments.

go-binary/templates/templates.go
tuunit: Ah my mistake, I was under the impression that sprig already has a toYaml function but they only have a toJson method and the PR for toYaml has been open for 3 years https://github.com/Masterminds/sprig/pull/360

go-binary/assets/config/configTypes.go
tuunit: Okay makes sense I was only thinking about the ingestion and not the generation part

Matthiator:
Heads-up: this PR currently conflicts with #221 (`feat/service-definition-catalog`) on core config/template files.

Reason:
- #221 migrates services to a dynamic catalog/map model (`Services map[string]ServiceInstance`, kebab-case keys, `index` access in templates).
- This PR still modifies the previous static service model (`GenericService`/`PersistentService`, camelCase paths).

Impact:
- Merge conflicts in `configTypes.go`, `factory.go`, related tests, and several customer `values.yaml.tplt` templates.
- Some changes are model-level incompatible (not only textual conflicts).

Proposal:
- First merge #221 as foundation.
- Then port the functional changes from this PR onto the catalog model:
  - move new service options into `configSchema` of ServiceDefinitions
  - adapt templates to `index .cluster.services "<service>" "config" ...`
  - update tests to dynamic services map / kebab-case keys.

tuunit (approved):

PR #221, opened 2026-04-15
feat: implement ServiceDefinition-based dynamic service registry

Matthiator:
Please read first: https://github.com/kubara-io/kubara/blob/main/CONTRIBUTING.md

## 📝 Summary
This PR implements a data-driven ServiceDefinition catalog and replaces hardcoded service wiring in kubara.

Key changes:
- Introduces `catalog/` package with loader for embedded + external ServiceDefinitions.
- Replaces static service typing with dynamic `Services map[string]ServiceInstance`.
- Converts built-in services to YAML ServiceDefinitions (`catalog/built-in/services/*.yaml`).
- Composes JSON schema for services dynamically from catalog definitions at runtime.
- Extracts default service config from `configSchema` defaults instead of hardcoded defaults.
- Migrates template access to dynamic map/index patterns with kebab-case service keys.
- Adds global `--catalog` and `--catalog-overwrite` flags for external catalog support.
- Introduces config versioning (`version: v1alpha1`) with a one-time migration for legacy unversioned configs that normalizes legacy camelCase keys and inline config, then rewrites the file. No runtime legacy support.
- Catalog loader rejects non-canonical service names with a hard error.
- Hardens catalog handling:
  - `clusterTypes` are now applied when deriving per-cluster defaults
  - strict `apiVersion` validation (`kubara.io/v1alpha1`)
  - `spec.status` is required on all ServiceDefinitions
  - no silent hardcoded fallback when catalog loading fails (errors are propagated)

Behavior note:
- Worker cluster service defaults are now `clusterTypes`-aware and may differ from previous implicit defaults.
- Legacy `config.yaml` files without a `version` field are automatically migrated and rewritten on first load.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

Executed locally:
- `cd go-binary && go test ./... -count=1`

## 🔗 Related Issues / Tickets
N/A

## ✅ Checklist


go-binary/catalog/built-in/services/argo-cd.yaml
tuunit, line 1: I think the service defines should be part of the templates directory

Lets take the opportunity and fully refactor the structure for this.

Move the catalog package to the toplevel in `go-binary/catalog/`

Instead of splitting up the `templates/embedded/**` from `catalog/services` move the contents of `templates/embedded/**` inside of the new `catalog/` into a directory called `built-in`.

Meaning `templates/embedded/` becomes `catalog/built-in/` and the built-in service definitions will be moved to `catalog/built-in/services/` as well

Furthermore, add a `catalog/builtin.go` that just exists to expose the `embed.FS`.
To be used inside the templating engine and catalog processing (will be further refactored in the future)

go-binary/catalog/built-in/services/argo-cd.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: kubara.io/v1alpha1
+kind: ServiceDefinition
+metadata:
+  name: argo-cd
```
tuunit, line 5: Lets already set the category annotation for all built-in services like on the homepage:

https://kubara.io/

```suggestion
  name: argo-cd
  annotations:
    kubara.io/category: application-management
```

go-binary/catalog/legacy_aliases.go
tuunit, line 1: lets call this `legacy_aliases.go`

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+	"embed"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	goYaml "go.yaml.in/yaml/v3"
```
tuunit:
```suggestion
	"go.yaml.in/yaml/v3"
```

tuunit: unnecessary custom dependency alias

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+
+	sort.Strings(files)
+	for _, path := range files {
+		content, err := fs.ReadFile(fsys, path)
+		if err != nil {
+			return Catalog{}, fmt.Errorf("read %q: %w", path, err)
+		}
+
+		var definition ServiceDefinition
+		if err := goYaml.Unmarshal(content, &definition); err != nil {
```
tuunit:
```suggestion
		if err := yaml.Unmarshal(content, &definition); err != nil {
```

go-binary/cmd/generate.go
```diff
@@ -162,13 +191,56 @@ func buildTemplateContext(clusterBlock config.Cluster, em envmap.EnvMap) (map[st
...
+		if configMap, ok := canonicalMap["config"].(map[string]any); ok {
+			for key, val := range configMap {
+				if _, exists := legacyService[key]; !exists {
+					legacyService[key] = val
+				}
+			}
+		}
+
+		services[legacyName] = legacyService
+	}
```
tuunit: do we really want this?

I think we should introduce a new version attribute to the config struct and set it to v1alpha1 by default. We can check if it doesn't exist in the config.yaml yet then we rewrite the config.yaml to the new format and convert the service names instead of supporting the old once

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+	goYaml "go.yaml.in/yaml/v3"
+)
+
+const servicesDirectory = "services"
+
+//go:embed services/*.yaml
+var embeddedServicesFS embed.FS
+
+type LoadOptions struct {
+	DistributionPath string
```
tuunit: we should make a decision now if we want to call it distributions or catalogs?

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+	builtIn, err := LoadBuiltIn()
+	if err != nil {
+		return Catalog{}, err
+	}
+
+	if strings.TrimSpace(options.DistributionPath) == "" {
+		return builtIn, nil
+	}
+
+	externalRoot, err := resolveExternalServicesPath(options.DistributionPath)
```
tuunit:
```suggestion
	externalRoot, err := resolveServicesPath(options.DistributionPath)
```

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+		if _, exists := merged.Services[name]; exists && !options.Overwrite {
+			return Catalog{}, fmt.Errorf("service definition %q already exists in built-in catalog", name)
+		}
+		merged.Services[name] = def
+	}
+
+	return merged, nil
+}
+
+func resolveExternalServicesPath(distributionPath string) (string, error) {
```
tuunit:
```suggestion
func resolveServicesPath(distributionPath string) (string, error) {
```

go-binary/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	external, err := loadFromFS(os.DirFS(externalRoot), ".")
+	if err != nil {
+		return Catalog{}, err
+	}
+
+	merged := builtIn.Clone()
+	for name, def := range external.Services {
+		if _, exists := merged.Services[name]; exists && !options.Overwrite {
```
tuunit, line 49: I don't like that we call it Overwrite in the options but the flag is just `--force`. We should really try to keep names consistent otherwise debugging will become annoying quite quickly.

So I propose either to call the flag `--overwrite` as well or `--force-overwrite` but that seems excessive

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+		return "", fmt.Errorf("stat distribution services path %q: %w", servicesDir, err)
+	}
+
+	rootInfo, err := os.Stat(cleaned)
+	if err != nil {
+		return "", fmt.Errorf("stat distribution path %q: %w", cleaned, err)
+	}
+	if !rootInfo.IsDir() {
+		return "", fmt.Errorf("distribution path %q is not a directory", cleaned)
+	}
```
tuunit: shouldn't the root be checked first?
and then the servicesDir?

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		return servicesDir, nil
+	}
+	if !errors.Is(err, os.ErrNotExist) {
+		return "", fmt.Errorf("stat distribution services path %q: %w", servicesDir, err)
+	}
+
+	rootInfo, err := os.Stat(cleaned)
+	if err != nil {
+		return "", fmt.Errorf("stat distribution path %q: %w", cleaned, err)
```
tuunit:
```suggestion
		return "", fmt.Errorf("catalog directory %q does not exist: %w", cleaned, err)
```

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if !rootInfo.IsDir() {
+		return "", fmt.Errorf("distribution path %q is not a directory", cleaned)
```
tuunit:
```suggestion
		return "", fmt.Errorf("catalog path %q is not a directory", cleaned)
```

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+
+	var files []string
+	if err := fs.WalkDir(fsys, root, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			return nil
+		}
+		if !strings.HasSuffix(strings.ToLower(path), ".yaml") {
```
tuunit: We should support yaml and yml

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+			return nil
+		}
+		files = append(files, path)
+		return nil
+	}); err != nil {
+		return Catalog{}, fmt.Errorf("walk service definitions: %w", err)
```
tuunit:
```suggestion
		return Catalog{}, fmt.Errorf("failed to walk service definitions: %w", err)
```

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+	sort.Strings(files)
+	for _, path := range files {
+		content, err := fs.ReadFile(fsys, path)
+		if err != nil {
+			return Catalog{}, fmt.Errorf("read %q: %w", path, err)
+		}
+
+		var definition ServiceDefinition
+		if err := goYaml.Unmarshal(content, &definition); err != nil {
+			return Catalog{}, fmt.Errorf("unmarshal %q: %w", path, err)
```
tuunit:
```suggestion
			return Catalog{}, fmt.Errorf("failed to unmarshal %q: %w", path, err)
```

go-binary/assets/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	}); err != nil {
+		return Catalog{}, fmt.Errorf("walk service definitions: %w", err)
+	}
+
+	sort.Strings(files)
+	for _, path := range files {
+		content, err := fs.ReadFile(fsys, path)
+		if err != nil {
+			return Catalog{}, fmt.Errorf("read %q: %w", path, err)
```
tuunit:
```suggestion
			return Catalog{}, fmt.Errorf("failed to read %q: %w", path, err)
```

go-binary/catalog/loader.go
```diff
@@ -0,0 +1,134 @@
...
+			return Catalog{}, fmt.Errorf("unmarshal %q: %w", path, err)
+		}
+		if err := definition.Validate(); err != nil {
+			return Catalog{}, fmt.Errorf("invalid service definition %q: %w", path, err)
+		}
+
+		canonicalName := CanonicalServiceName(definition.Metadata.Name)
+		if canonicalName != definition.Metadata.Name {
+			definition.Metadata.Name = canonicalName
+		}
```
tuunit: This is only relevant for the legacy names correct?

go-binary/catalog/types.go
```diff
@@ -0,0 +1,139 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+type Status string
+
+const (
+	ServiceDefinitionAPIVersion        = "kubara.io/v1alpha1"
+	StatusEnabled               Status = "enabled"
+	StatusDisabled              Status = "disabled"
+)
```
tuunit:
```suggestion
type Status string

const (
	StatusEnabled               Status = "enabled"
	StatusDisabled              Status = "disabled"
)

const (
	ServiceDefinitionAPIVersion string = "kubara.io/v1alpha1"
)
```

we should treat the first one as an enum and therefore i would suggest to split the const

go-binary/assets/catalog/types.go
```diff
@@ -0,0 +1,139 @@
...
+type Metadata struct {
+	Name        string            `yaml:"name"`
+	Annotations map[string]string `yaml:"annotations,omitempty"`
+}
+
+type ServiceSpec struct {
+	ChartPath    string                           `yaml:"chartPath"`
+	AppName      string                           `yaml:"appName,omitempty"`
+	Default      Status                           `yaml:"default,omitempty"`
+	Status       Status                           `yaml:"status,omitempty"` // compatibility with early proposal drafts
```
tuunit:
```suggestion
	Status       Status                           `yaml:"status,omitempty"`
```

remove the default field. we should always require the field

go-binary/catalog/types.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (s *ServiceSpec) UnmarshalYAML(value *goYaml.Node) error {
+	type serviceSpecAlias struct {
+		ChartPath    string         `yaml:"chartPath"`
+		AppName      string         `yaml:"appName,omitempty"`
+		Default      Status         `yaml:"default,omitempty"`
+		Status       Status         `yaml:"status,omitempty"`
+		ClusterTypes []string       `yaml:"clusterTypes,omitempty"`
+		ConfigSchema map[string]any `yaml:"configSchema,omitempty"`
+	}
```
tuunit: why this anonymous struct?

tuunit: why not unmarshal directly into ServiceSpec?
I guess because the tags are different between json vs yaml 🤔 
But as far as I known https://github.com/goccy/go-yaml supports the json tags as well maybe we need to specify all structs in the catalog types with json to make it work 🤔

go-binary/assets/catalog/types.go
```diff
@@ -0,0 +1,139 @@
+package catalog
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	goYaml "go.yaml.in/yaml/v3"
```
tuunit:
```suggestion
	"go.yaml.in/yaml/v3"
```

go-binary/assets/catalog/types.go
```diff
@@ -0,0 +1,139 @@
...
+type ServiceSpec struct {
+	ChartPath    string                           `yaml:"chartPath"`
+	AppName      string                           `yaml:"appName,omitempty"`
+	Default      Status                           `yaml:"default,omitempty"`
+	Status       Status                           `yaml:"status,omitempty"` // compatibility with early proposal drafts
+	ClusterTypes []string                         `yaml:"clusterTypes,omitempty"`
+	ConfigSchema *apiextensionsv1.JSONSchemaProps `yaml:"configSchema,omitempty"`
+}
+
+func (s *ServiceSpec) UnmarshalYAML(value *goYaml.Node) error {
```
tuunit:
```suggestion
func (s *ServiceSpec) UnmarshalYAML(value *yaml.Node) error {
```

go-binary/catalog/types.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return fmt.Errorf("marshal configSchema: %w", err)
+	}
+
+	var schema apiextensionsv1.JSONSchemaProps
+	if err := json.Unmarshal(configSchemaBytes, &schema); err != nil {
+		return fmt.Errorf("unmarshal configSchema into JSONSchemaProps: %w", err)
+	}
+
+	s.ConfigSchema = &schema
```
tuunit: I think we can cut this down to one unmarshal

go-binary/catalog/types.go
tuunit, line 1: Please add godocs (comments) to all the fields in this file

go-binary/assets/catalog/types.go
```diff
@@ -0,0 +1,139 @@
...
+func (s ServiceSpec) EffectiveDefaultStatus() Status {
+	switch {
+	case s.Default == StatusEnabled || s.Default == StatusDisabled:
+		return s.Default
+	case s.Status == StatusEnabled || s.Status == StatusDisabled:
+		return s.Status
+	default:
+		return StatusDisabled
+	}
+}
```
tuunit: I think this should be removed as outlined above we should always require the status field to be set and it should fail with a validation / unmarshal error when the field is missing from a service. it is best practices to always have the negative / false state as the default for a variable type

go-binary/assets/config/configMngr.go
```diff
@@ -8,62 +8,84 @@ import (
...
+	"sort"
+
+	"kubara/assets/catalog"
 
 	"github.com/go-viper/mapstructure/v2"
 	"github.com/invopop/jsonschema"
-	"github.com/knadh/koanf/providers/file"
-	"github.com/knadh/koanf/v2"
 	schemaValidator "github.com/santhosh-tekuri/jsonschema/v6"
 	goYaml "go.yaml.in/yaml/v3"
```
tuunit:
```suggestion
	"go.yaml.in/yaml/v3"
```

go-binary/assets/config/configMngr.go
```diff
@@ -8,62 +8,84 @@ import (
...
-	if err := k.Load(file.Provider(cm.filepath), yaml.Parser()); err != nil {
-		return fmt.Errorf("failed to load config file: %w", err)
+	data, err := os.ReadFile(cm.filepath)
+	if err != nil {
+		return fmt.Errorf("failed to read config file: %w", err)
 	}
 
-	// Decoderconfig is used to handle mapping between koanf variables and structs with inline fields when reading from file
+	var raw map[string]any
+	if err := goYaml.Unmarshal(data, &raw); err != nil {
```
tuunit:
```suggestion
	if err := yaml.Unmarshal(data, &raw); err != nil {
```

go-binary/assets/config/configMngr.go
```diff
@@ -8,62 +8,84 @@ import (
...
+		return fmt.Errorf("failed to parse yaml config: %w", err)
+	}
+
+	decodeHook := mapstructure.DecodeHookFuncType(func(from reflect.Type, to reflect.Type, source any) (any, error) {
+		if to != reflect.TypeFor[ServiceInstance]() {
+			return source, nil
+		}
+		return decodeServiceInstance(source)
+	})
```
tuunit:
```suggestion
	func decodeServiceHook() mapstructure.DecodeHookFunc {
		return func(_ reflect.Type, to reflect.Type, data any) (any, error) {
			if to != reflect.TypeFor[ServiceInstance]() {
				return source, nil
			}
			return decodeServiceInstance(data)
		}
	}
```

please put this into the `services.go` file as well

go-binary/assets/config/configMngr.go
```diff
@@ -8,62 +8,84 @@ import (
...
 	dc := &mapstructure.DecoderConfig{
 		TagName:          "yaml",
 		WeaklyTypedInput: false,
 		Result:           cm.config,
 		Squash:           true,
+		DecodeHook:       decodeHook,
```
tuunit:
```suggestion
		DecodeHook:       mapstructure.ComposeDecodeHookFunc(
			decodeServiceHook
		),
```

tuunit: for future extendability

go-binary/assets/config/configMngr.go
```diff
@@ -87,11 +109,23 @@ func GenerateSchema() (map[string]any, error) {
 		return nil, fmt.Errorf("unmarshal schema: %w", err)
 	}
 
+	cat, err := catalog.Load(catalogOptions)
+	if err != nil {
+		return nil, fmt.Errorf("load service catalog: %w", err)
```
tuunit:
```suggestion
		return nil, fmt.Errorf("failed loading catalog: %w", err)
```

go-binary/assets/config/configMngr.go
```diff
@@ -161,3 +195,74 @@ func (cm *Manager) SaveToFile() error {
 
 	return nil
 }
+
+func composeServiceSchema(schemaDoc map[string]any, cat catalog.Catalog) error {
+	defs, ok := schemaDoc["$defs"].(map[string]any)
+	if !ok {
+		return fmt.Errorf("schema missing $defs")
```
tuunit:
```suggestion
		return fmt.Errorf("catalog schema is missing $defs")
```

tuunit: this is actually required? generated automatically?

go-binary/assets/config/configMngr.go
```diff
@@ -161,3 +195,74 @@ func (cm *Manager) SaveToFile() error {
...
+	}
+	sort.Strings(keys)
+
+	serviceProperties := make(map[string]any, len(keys))
+	required := make([]any, 0, len(keys))
+	for _, serviceName := range keys {
+		definition := cat.Services[serviceName]
+		instanceSchema, err := buildServiceInstanceSchema(definition)
+		if err != nil {
+			return nil, fmt.Errorf("build schema for service %q: %w", serviceName, err)
```
tuunit:
```suggestion
			return nil, fmt.Errorf("failed to build schema for service %q: %w", serviceName, err)
```

go-binary/assets/config/configMngr.go
```diff
@@ -161,3 +195,74 @@ func (cm *Manager) SaveToFile() error {
...
+
+func buildServiceInstanceSchema(definition catalog.ServiceDefinition) (map[string]any, error) {
+	defaultStatus := string(toConfigStatus(definition.Spec.EffectiveDefaultStatus()))
+	properties := map[string]any{
+		"status": map[string]any{
+			"type":        "string",
+			"title":       "Service Status",
+			"description": "The desired status of the service.",
+			"enum":        []any{string(StatusEnabled), string(StatusDisabled)},
+			"default":     defaultStatus,
```
tuunit:
```suggestion
```

go-binary/assets/config/configMngr.go
```diff
@@ -161,3 +195,74 @@ func (cm *Manager) SaveToFile() error {
...
+		"title":                "Services",
+		"description":          "Configuration for deployed services.",
+		"additionalProperties": false,
+		"properties":           serviceProperties,
+		"required":             required,
+	}, nil
+}
+
+func buildServiceInstanceSchema(definition catalog.ServiceDefinition) (map[string]any, error) {
+	defaultStatus := string(toConfigStatus(definition.Spec.EffectiveDefaultStatus()))
```
tuunit:
```suggestion
```

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
+package config
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"kubara/assets/catalog"
+
+	goYaml "go.yaml.in/yaml/v3"
```
tuunit:
```suggestion
	"go.yaml.in/yaml/v3"
```

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+
+		instance := ServiceInstance{
+			Status: toConfigStatus(status),
+			Config: extractDefaultsFromSchema(def.Spec.ConfigSchema),
+		}
+		out[name] = instance
+	}
+	return out
+}
```
tuunit:
```suggestion
func createServicesFromCatalogDefaults(cat catalog.Catalog) Services {
	out := make(Services, len(cat.Services))
	for name, def := range cat.Services {
		instance := ServiceInstance{
			Status: catalog.StatusDisabled,
			Config: extractDefaultsFromSchema(def.Spec.ConfigSchema),
		}
		out[name] = instance
	}
	return out
}
```

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+func newDefaultServicesFromCatalogWithOptions(options catalog.LoadOptions, clusterType string) (Services, error) {
+	cat, err := catalog.Load(options)
+	if err != nil {
+		return nil, err
+	}
+	return serviceDefaultsFromCatalog(cat, clusterType), nil
+}
```
tuunit:
```suggestion
func createServicesFromCatalogWithOptions(options catalog.LoadOptions, clusterType string) (Services, error) {
	cat, err := catalog.Load(options)
	if err != nil {
		return nil, err
	}
	return createServicesFromCatalogDefaults(cat), nil
}
```

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
...
+		instance := input[key]
+		if existing, ok := normalized[canonical]; ok {
+			normalized[canonical] = mergeServiceInstances(existing, instance)
+			continue
+		}
+		normalized[canonical] = instance
+	}
+
+	return normalized
+}
```
tuunit: I guess all of this would be removed with preempting the legacy conversion and forcing `config.yaml`s into the new structure

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
...
+	out := make(map[string]any, len(in))
+	for k, v := range in {
+		if nested, ok := v.(map[string]any); ok {
+			out[k] = cloneMap(nested)
+			continue
+		}
+		out[k] = v
+	}
+	return out
+}
```
tuunit: we need to refactor the `utils.go` as that is an anti pattern and we don't just want to have a general dump for stuff but nevertheless best to move this generic function their until we have a better place

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
...
+				return true
+			}
+		}
+		return false
+	default:
+		// Unknown/placeholder cluster types keep the legacy behavior:
+		// expose all services with their definition defaults.
+		return true
+	}
+}
```
tuunit:
```suggestion
```

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
...
+	for key, val := range raw {
+		if key == "status" || key == "config" {
+			continue
+		}
+		legacyInline[key] = val
+	}
+	if len(legacyInline) > 0 {
+		// Explicit "config" key should win over legacy inline keys when both exist.
+		out.Config = mergeConfigDefaults(legacyInline, out.Config)
+	}
```
tuunit:
```suggestion
```

go-binary/assets/config/services.go
```diff
@@ -0,0 +1,329 @@
...
+	case []any:
+		out := make([]any, len(typed))
+		for i := range typed {
+			out[i] = normalizeValue(typed[i])
+		}
+		return out
+	default:
+		return v
+	}
+}
```
tuunit: move to utils as well for now. we should probably have a package for conversions or cloning

go-binary/cmd/bootstrap.go
```diff
@@ -211,6 +227,19 @@ func (flags *BootstrapFlags) AddFlags(cmd *cli.Command) {
...
 		},
+		&cli.StringFlag{
+			Name:        "catalog",
+			Value:       flags.CatalogPath,
+			Usage:       "Path to external ServiceDefinition catalog/distribution directory.",
+			Destination: &flags.CatalogPath,
+		},
+		&cli.BoolFlag{
+			Name:        "force",
+			Aliases:     []string{"overwrite"},
```
tuunit:
```suggestion
			Name:        "overwrite",
```

go-binary/cmd/generate.go
```diff
@@ -87,6 +94,13 @@ func (flags *GenerateFlags) ToOptions(cmd *cli.Command) (*GenerateOptions, error
 	if err != nil {
 		return nil, fmt.Errorf("failed to get overlay values path: %w", err)
 	}
+	catalogPath := ""
+	if flags.CatalogPath != "" {
+		catalogPath, err = utils.GetFullPath(cmd.String("catalog"), cwd)
```
tuunit: why use flags.CatalogPath in the condition but then `cmd.String("catalog")` in the path retrieval?

go-binary/cmd/bootstrap.go
```diff
@@ -24,6 +25,8 @@ type BootstrapFlags struct {
 	ClusterSecretStorePath string
 	ManagedCatalogPath     string
 	OverlayValuesPath      string
+	CatalogPath            string
+	CatalogOverwrite       bool
```
tuunit: Those should be global flags in `main.go` and removed from all other commands

go-binary/cmd/generate.go
```diff
@@ -162,13 +191,56 @@ func buildTemplateContext(clusterBlock config.Cluster, em envmap.EnvMap) (map[st
 	if err := json.Unmarshal(clusterJSON, &clusterMap); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal clusterJSON to map: %w", err)
 	}
+	addLegacyServiceAliases(clusterMap)
```
tuunit: instead of supporting them. this would be the place to enforce an update to the config.yaml

tuunit (changes requested):

go-binary/catalog/built-in/services/argo-cd.yaml
Matthiator: Done. `catalog/` is now a top-level package with `built-in/` containing both service definitions and embedded templates. [`catalog/builtin.go`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/catalog/builtin.go) exposes the `embed.FS`, consumed by the template engine in [`templates/templates.go:32`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/templates/templates.go#L32).

go-binary/cmd/generate.go
Matthiator: Done. Config has a `version` field ([`configTypes.go:5`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/configTypes.go#L5)). A single version-gated `migrateRawConfig()` in [`migration.go`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/migration.go) normalizes legacy service keys and inline config, sets `version: v1alpha1`, and rewrites the file. No runtime legacy support remains — `decodeServiceInstance()` in [`services.go`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/services.go#L213) only reads `status` and `config`.

go-binary/assets/catalog/loader.go
Matthiator: Decided: `catalog` consistently. `LoadOptions.CatalogPath`, `--catalog` flag, all error messages. "Distribution" implies packaging/shipping which is still out-of-scope.

go-binary/catalog/loader.go
Matthiator: Updated for final naming: the global catalog collision flag is --catalog-overwrite (mapped to LoadOptions.Overwrite), and --force is removed.

go-binary/catalog/loader.go
Matthiator: Correct — and we took it further: the loader now **rejects** non-canonical names with a hard error instead of silently converting them ([`catalog/loader.go:122-127`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/catalog/loader.go#L122-L127)). ServiceDefinitions are a new format and must use canonical kebab-case names. Legacy name normalization only happens in the config migration ([`migration.go:89`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/migration.go#L89)).

go-binary/catalog/types.go
Matthiator: Replaced with named `serviceSpecYAML` struct ([`catalog/types.go:58-64`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/catalog/types.go#L58-L64)).

Matthiator: Because `ConfigSchema` is `apiextensionsv1.JSONSchemaProps` which uses JSON struct tags, not YAML tags. Direct YAML unmarshal doesn't populate it correctly. We now use `runtime.DefaultUnstructuredConverter` ([`catalog/types.go:83`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/catalog/types.go#L83)) instead of the previous JSON roundtrip — the idiomatic K8s approach for this conversion.

go-binary/catalog/types.go
Matthiator: Done. JSON marshal/unmarshal roundtrip replaced with `runtime.DefaultUnstructuredConverter.FromUnstructured()` ([`catalog/types.go:83`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/catalog/types.go#L83)). Single YAML unmarshal into the intermediate struct, then K8s runtime conversion for the schema.

go-binary/assets/catalog/types.go
Matthiator: Done. `spec.status` is required and validated in `Validate()` ([`catalog/types.go:122-124`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/catalog/types.go#L122-L124)). `EffectiveDefaultStatus()` is removed. The `default` field in the JSON Schema status property was also removed per your deletion suggestions.

go-binary/assets/config/configMngr.go
Matthiator: Yes. `$defs` is generated automatically by the `jsonschema.Reflector`. [`composeServiceSchema()`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/configMngr.go#L208) then replaces the `Services` entry within `$defs` with the dynamically composed service schemas from the catalog.

go-binary/assets/config/services.go
Matthiator: Good point. We intentionally keep `spec.status` as the source of truth and apply `clusterTypes` filtering on top (see `services.go:80-95`).

If we defaulted all services to `disabled` here, we would ignore explicit defaults from ServiceDefinitions (for example `status: enabled`) and duplicate policy outside the catalog.

go-binary/assets/config/services.go
Matthiator: Done. `normalizeServiceKeys()` is removed from `services.go`. Key normalization now happens in the migration ([`normalizeRawServiceKeys()` in `migration.go:87`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/migration.go#L87)), not at runtime.

go-binary/assets/config/services.go
Matthiator: Done. `toStringMap` and `normalizeValue` moved to [`utils.go`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/utils.go). `cloneMap` still in `services.go` — can be moved when a proper conversions package is introduced.

go-binary/cmd/generate.go
Matthiator: Done. Centralized via [`catalogLoadOptionsFromCommand()`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/cmd/catalog_flags.go#L14) in `cmd/catalog_flags.go`. Single consistent path for all commands.

go-binary/cmd/bootstrap.go
Matthiator: Updated for final naming: --catalog and --catalog-overwrite are global flags in main.go; init keeps its own command-local --overwrite for config overwrite.

go-binary/cmd/generate.go
Matthiator: Done. Config migration in [`migration.go`](https://github.com/kubara-io/kubara/blob/e8177641c9cb7ecf7e7f82fc5bcfdfe21bec0547/go-binary/assets/config/migration.go) rewrites the file on first load — normalizes keys, moves inline config under `config` sub-key, sets `version: v1alpha1`. No runtime legacy support.

go-binary/catalog/built-in/services/argo-cd.yaml
Matthiator: Done.

go-binary/catalog/legacy_aliases.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/assets/catalog/loader.go
Matthiator: Done.

go-binary/catalog/types.go
Matthiator: Done.

go-binary/assets/catalog/types.go
Matthiator: Done.

go-binary/assets/catalog/types.go
Matthiator: Done.

go-binary/assets/catalog/types.go
Matthiator: Done.

go-binary/catalog/types.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/configMngr.go
Matthiator: Done.

go-binary/assets/config/services.go
Matthiator: Done.

go-binary/assets/config/services.go
Matthiator: Done.

go-binary/assets/config/services.go
Matthiator: Done.

go-binary/assets/config/services.go
Matthiator: Done.

go-binary/assets/config/services.go
Matthiator: Done.

go-binary/cmd/bootstrap.go
Matthiator: Done.

go-binary/catalog/loader.go
Matthiator: Quick follow-up to keep this thread accurate: we adjusted naming once more to avoid collision with init's command-local `--overwrite`.

Final state is global `--catalog-overwrite` (mapped to `LoadOptions.Overwrite`) plus command-local init `--overwrite` for config overwrite. So the old `--force` mismatch is gone, and semantics are now explicit.

go-binary/assets/config/services.go
Matthiator: Adding context to this thread: the suggestion block itself was empty, so there was no concrete code snippet to apply.

We still simplified this area in the final code path (cluster type applicability + strict service decoding/migration split), and kept the implementation aligned with the surrounding review comments.

go-binary/catalog/types.go
tuunit: @Matthiator but why doesn't it properly unmarshal it? the library supposedly supports json tags

Matthiator: Good question.

You’re right that JSON-tag reuse works in the YAML->JSON wrapper approach (`sigs.k8s.io/yaml`).
In this code path, however, we decode directly with `go.yaml.in/yaml/v3`, and JSON-tag fallback is not a stable contract (historical/ongoing discussion: https://github.com/go-yaml/yaml/issues/424 and https://github.com/yaml/go-yaml/issues/153).

The bigger blocker here is `apiextensionsv1.JSONSchemaProps` itself: fields like `default`/`enum` use `v1.JSON`.
Direct YAML decode into `JSONSchemaProps` fails for scalar defaults (e.g. `cannot unmarshal !!str/!!int into v1.JSON`).

That is why we decode `configSchema` into `map[string]any` first and then convert with
`runtime.DefaultUnstructuredConverter.FromUnstructured(...)`.
This follows the Kubernetes unstructured->typed conversion path and correctly materializes `JSONSchemaProps` (including scalar defaults).

tuunit (approved):

PR #229, opened 2026-04-16
fix: argo out-off-sync issues because of newly introduced external secret null byte policy

tuunit:
## 📝 Summary
All Argo CD apps have been constantly out-of-sync because external-secrets introduced a new field called nullBytePolicy for how to handle secrets that have no contents.

This is now being set to "Fail" with this PR. To indicate to the user that the contents of a secret will not be of use to an application, to more easily trace the issue and not allow applications to be able to start as they expect an mountable secret to have valid content.

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [X] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [ ] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [X] CI passed
- [ ] Unit tested
  
## ✅ Checklist
  
https://external-secrets.io/main/api/spec/#external-secrets.io/v1.ExternalSecretNullBytePolicy
https://github.com/external-secrets/external-secrets/pull/6194
https://github.com/external-secrets/external-secrets/releases/tag/v2.3.0

### DIFF
--- a/go-binary/templates/embedded/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecret.es.argo.cluster.tpl
+++ b/go-binary/templates/embedded/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecret.es.argo.cluster.tpl
@@ -44,5 +44,6 @@ spec:
         conversionStrategy: Default
         decodingStrategy: None
         metadataPolicy: None
+        nullBytePolicy: Fail
 ---
 {{- end }}
--- a/go-binary/templates/embedded/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecret.es.argo.repo.tpl
+++ b/go-binary/templates/embedded/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecret.es.argo.repo.tpl
@@ -50,5 +50,6 @@ spec:
         conversionStrategy: Default
         decodingStrategy: None
         metadataPolicy: None
+        nullBytePolicy: Fail
 ---
 {{- end }}
--- a/go-binary/templates/embedded/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.ces.dockerPullSecret.tpl
+++ b/go-binary/templates/embedded/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.ces.dockerPullSecret.tpl
@@ -39,5 +39,6 @@ spec:
           conversionStrategy: Default
           decodingStrategy: None
           metadataPolicy: None
+          nullBytePolicy: Fail
 ---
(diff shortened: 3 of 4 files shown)


tuunit:
Helm linting is failing because of outdated CRDs catalog for the schema validation

https://github.com/datreeio/CRDs-catalog/pull/849

I think we shouldn't rely on that repository and instead use the sources from the original CRDs

Matthiator:
Suggestion: could we set `nullBytePolicy: Ignore` in this PR instead of `Fail`?

`Ignore` still resolves the Argo CD out-of-sync drift by making the field explicit, while preserving current runtime behavior. `Fail` introduces stricter semantics and may break existing workloads if secrets contain NUL bytes. If we want `Fail`, it would be safer as a separate hardening change with docs/migration notes.

Matthiator:
@tuunit Agree. The current failure is caused by the outdated schema in `datreeio/CRDs-catalog` (it does not include `nullBytePolicy` yet).

Let’s keep this PR focused on the `nullBytePolicy` change and open a follow-up PR to switch Helm schema validation to upstream CRDs (or a pinned schema source derived from upstream CRDs), so CI is no longer blocked by third-party catalog lag.

Matthiator:
Tracking issue created: #230

tuunit:
> Suggestion: could we set `nullBytePolicy: Ignore` in this PR instead of `Fail`?
> 
> `Ignore` still resolves the Argo CD out-of-sync drift by making the field explicit, while preserving current runtime behavior. `Fail` introduces stricter semantics and may break existing workloads if secrets contain NUL bytes. If we want `Fail`, it would be safer as a separate hardening change with docs/migration notes.

I would argue against Ignore. I've seen it first hand how annoying it is to debug a failing not properly working application just to realise the contents of the secret have been wrong all along therefore I would rather have external-secrets show me an error on the ES resource and not create a k8s secret with none functional contents. The sooner the error is being "displayed" / discoverable up the chain the better.

What is better:

Application in weird crash loop and you have to figure out why because it might not be obvious that a secret environment variable or file is just empty

Or:

Application doesn't even try to just as k8s will tell you secret is missing and External-Secrets shows an error on the ES resource that it cannot create a secret in k8s because the source doesn't have any contents. (I argue this is way cleaner and easier to identify)

fabian-schmitt (commented):
Making sure that the secret isn't empty seems to be the most straight-forward solution for me.
So i tend to think "fail" is the better solution.

tuunit:
As discussed internally we decided to use Fail as it makes it easier to identify the problematic secrets higher in the chain and should lead to shorter debugging sessions in case of issues

PR #284, opened 2026-05-06
feat: automatic config migration from legacy definition to schema driven approach

tuunit:
## 📝 Summary
Automatic migration of legacy configs into the new schema driven catalog service definitions.

## 🧩 Type of change
- [X] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [X] 🧪 Test or CI change
- [X] ♻️ Refactor / cleanup

## 🧪 Testing
- [X] CI passed
- [X] Unit tested
  
## ✅ Checklist

### DIFF
--- a/src/cmd/bootstrap.go
+++ b/src/cmd/bootstrap.go
@@ -130,9 +130,6 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
 	if err := cs.Load(); err != nil {
 		return nil, fmt.Errorf("load config: %w", err)
 	}
-	if err := cs.Validate(); err != nil {
-		return nil, fmt.Errorf("validate config: %w", err)
-	}
 
 	// Find the cluster by name from the argument
 	clusterName := cmd.StringArg("cluster-name")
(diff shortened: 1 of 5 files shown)


Matthiator (changes requested):
Looks good to me overall.

One thing I would consider before merging: can we validate the migrated config before writing it back to disk? Right now `Load()` migrates, applies defaults, and immediately calls `SaveToFile()`. The command-level `Validate()` calls in `generate` / `bootstrap` happen only afterwards, so an invalid legacy config could already be rewritten before the command fails.

Suggested flow would be: migrate -> decode/apply defaults -> validate migrated config -> persist. That way we only modify the user config once we know the migrated result is valid.

tuunit:
> Looks good to me overall.
> 
> One thing I would consider before merging: can we validate the migrated config before writing it back to disk? Right now `Load()` migrates, applies defaults, and immediately calls `SaveToFile()`. The command-level `Validate()` calls in `generate` / `bootstrap` happen only afterwards, so an invalid legacy config could already be rewritten before the command fails.
> 
> Suggested flow would be: migrate -> decode/apply defaults -> validate migrated config -> persist. That way we only modify the user config once we know the migrated result is valid.

Good point. I decided to move the validation inside the load method and make it a private member function as a config load without validation doesn't make sense anyways.

Matthiator (approved):

PR #317, opened 2026-05-17
chore(renovate): group controller-runtime with k8s modules

Matthiator:
## 📝 Summary
Adds an explicit Renovate override for `sigs.k8s.io/controller-runtime` so it uses the existing `k8s.io` Go module group while keeping the generic `sigs.k8s.io` group unchanged.

`controller-runtime` needs to stay aligned with Kubernetes/client-go minor versions. This avoids broken Renovate PRs like #311 where `k8s.io/*` was updated without the compatible `controller-runtime` version.

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [x] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [ ] Unit tested
- [x] Not tested (explain why below)

No runtime tests were run because this is a Renovate configuration-only change. The config file was validated as JSON with `jq empty renovate.json`.

## 🔗 Related Issues / Tickets
Related to #311

## ✅ Checklist

Renovate's built-in Kubernetes monorepo preset covers the `kubernetes/kubernetes` staging repositories (`k8s.io/*`) but does not include `sigs.k8s.io/controller-runtime`. The preset was introduced in renovatebot/renovate#41171 and consolidated in renovatebot/renovate#41454.

`controller-runtime` is still tied to the Kubernetes/client-go compatibility matrix: https://github.com/kubernetes-sigs/controller-runtime#compatibility

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -49,12 +49,21 @@
       "groupSlug": "k8s-io-go-modules"
     },
     {
-      "description": "Group sigs.k8s.io go modules (minor and patch)",
+      "description": "Group sigs.k8s.io go modules except controller-runtime (minor and patch)",
       "matchManagers": ["gomod"],
-      "matchPackagePrefixes": ["sigs.k8s.io/"],
+      "matchPackageNames": ["sigs.k8s.io/**", "!sigs.k8s.io/controller-runtime"],
       "matchUpdateTypes": ["minor", "patch"],
       "groupName": "sigs.k8s.io go modules",
       "groupSlug": "sigs-k8s-io-go-modules"
+    },
+    {
+      "description": "Group controller-runtime with k8s.io go modules",
+      "matchManagers": ["gomod"],
+      "matchPackageNames": ["sigs.k8s.io/controller-runtime"],
+      "matchUpdateTypes": ["minor", "patch"],
+      "enabled": true,
+      "groupName": "k8s.io go modules",
+      "groupSlug": "k8s-io-go-modules"
     },
     {
       "description": "Group golang.org/x go modules (minor and patch)",
--- a/renovate.json
+++ b/renovate.json
@@ -49,15 +49,15 @@
       "groupSlug": "k8s-io-go-modules"
     },
     {
-      "description": "Group sigs.k8s.io go modules except controller-runtime (minor and patch)",
+      "description": "Group sigs.k8s.io go modules (minor and patch)",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -37,13 +37,22 @@
       "groupSlug": "k8s-io-go-modules"
     },
     {
-      "description": "Group sigs.k8s.io go modules (minor and patch)",
+      "description": "Group sigs.k8s.io go modules except controller-runtime (minor and patch)",
       "matchManagers": ["gomod"],
-      "matchPackagePrefixes": ["sigs.k8s.io/"],
+      "matchPackageNames": ["sigs.k8s.io/**", "!sigs.k8s.io/controller-runtime"],
```
tuunit: Mhhh does this really make sense 🤔 
Won't this lead to a version drift in the longterm? Why wasn't this caught in CI?

Matthiator: Good point. I checked Renovate upstream as well. Renovate now has built-in Kubernetes monorepo presets (`monorepo:kubernetes` / `group:kubernetesMonorepo`), introduced via [renovatebot/renovate#41171](https://github.com/renovatebot/renovate/pull/41171) and consolidated/renamed via [renovatebot/renovate#41454](https://github.com/renovatebot/renovate/pull/41454).

That preset is generated from the `kubernetes/kubernetes` staging repos, so it intentionally covers `k8s.io/*` modules only. `sigs.k8s.io/controller-runtime` is not included by design, although it still follows the [controller-runtime / client-go compatibility matrix](https://github.com/kubernetes-sigs/controller-runtime#compatibility).

So this is not really a Renovate bug; it is a gap between Renovate’s Kubernetes monorepo definition and controller-runtime’s compatibility relationship. We hit that in [#311](https://github.com/kubara-io/kubara/pull/311), where the `k8s.io/*` update needed `controller-runtime` to move with it.

CI did not catch this because our checks validate build/tests, not the upstream support matrix; an unsupported pair can still compile.

I’ll revise our first approach and switch this to an explicit override instead, so the generic `sigs.k8s.io` group stays intact while the `controller-runtime` exception is visible. Commit follows.

tuunit (approved):

PR #321, opened 2026-05-18
chore: right-size managed service resources

Matthiator:
## 📝 Summary

  
Right-sizes resource requests and memory limits for managed service Helm charts based on observed dev cluster usage and KRR recommendations.

This reduces excessive CPU requests, especially for Argo CD and Alloy, keeps CPU limits absent from rendered workloads, and aligns memory requests with memory limits for more predictable scheduling and capacity planning. Grafana memory is increased to account for observed OOMKills during dashboard usage.

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [x] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [ ] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

Deploy testing is still pending; this is opened as a draft PR for validation on the target cluster.
  
## 🔗 Related Issues / Tickets
N/A

## ✅ Checklist
  
Validated with local Helm rendering across the managed service catalog.

Checks performed:
- `go test ./...`
- YAML parse for changed values files
- `git diff --check`
- rendered resource scan: no `resources.limits.cpu`
- rendered resource scan: memory requests match memory limits where memory resources are set

Resource changes cover Argo CD, Loki/Alloy, kube-prometheus-stack/Grafana, Traefik, Kyverno, and Kyverno Policy Reporter.

### DIFF
--- a/go-binary/templates/embedded/managed-service-catalog/helm/argo-cd/values.yaml
+++ b/go-binary/templates/embedded/managed-service-catalog/helm/argo-cd/values.yaml
@@ -33,9 +33,9 @@ argo-cd:
     resources:
       requests:
         cpu: 50m
-        memory: 150Mi
+        memory: 256Mi
       limits:
-        memory: 200Mi
+        memory: 256Mi
     livenessProbe:
       enabled: true
     readinessProbe:
@@ -46,7 +46,7 @@ argo-cd:
         cpu: 100m
         memory: 256Mi
       limits:
-        memory: 512Mi
+        memory: 256Mi
     livenessProbe:
       enabled: true
     readinessProbe:
@@ -60,10 +60,10 @@ argo-cd:
       enabled: false
     resources:
       requests:
-        cpu: 256m
-        memory: 500Mi
+        cpu: 100m
+        memory: 512Mi
       limits:
-        memory: 1000Mi
+        memory: 512Mi
     livenessProbe:
       enabled: true
     readinessProbe:
@@ -81,10 +81,10 @@ argo-cd:
       automountServiceAccountToken: true
     resources:
       requests:
-        cpu: 500m
+        cpu: 100m
         memory: 512Mi
       limits:
-        memory: 1000Mi
+        memory: 512Mi
   configs:
     cm:
       exec.enabled: false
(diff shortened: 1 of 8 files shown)


tuunit (commented):
I would not decrease the limits for the argo cd components. only the requests

Matthiator:
> I would not decrease the limits for the argo cd components. only the requests

We can also make this PR more conservative by only changing CPU requests and keeping the previous memory limits, especially for the Argo CD components. The CPU request reduction is the main part that improves scheduling headroom; memory right-sizing is a separate tradeoff.

My preference would be to keep request=limit for predictable capacity planning, since the values were checked with KRR and bootstrapping worked reliably in our test cluster. But if we want to minimize rollout risk, I am also fine with restoring the previous Argo CD memory limits and revisiting memory sizing separately.

tuunit:
As we want to support people who might not have the right team size or experience I don't think it makes sense to have requests=limit for the memory footprint as this will cause overutilization for a lot of people with smaller workloads but for teams with bigger loads it will hit the limits quite quickly and therefore lead to unexpected crashes.

In my opinion, requests should reflect the idle default state and limits should support spikes / bursts.

I don't want our components to always be overutilized but still support a little bit extra load for the first couple of workloads a customer might add

Matthiator:
We discussed this internally in the Kubara team and agreed to go with the current approach for now.

The intention behind setting memory `requests = limits` for the managed Kubara components is to provide predictable bootstrap defaults that worked reliably in our test/dev setup, with some buffer. This should make scheduling and capacity planning more predictable and avoid starting with overly optimistic memory requests.

At the same time, we should treat these values as initial Kubara defaults, not as final values for every cluster forever. Once teams add their own workloads, they will need to revisit resource sizing based on actual usage, alerts, OOMKills, or workload growth. That periodic tuning is expected.

So for this PR, the team agreement is:

- keep memory `requests = limits` as the initial default/best-practice for the affected managed components
- keep avoiding CPU limits
- consider the concrete values bootstrap defaults with buffer
- observe how this behaves in practice and revisit the defaults if user feedback or operational data suggests a better approach

This is a pragmatic team decision for now, and we can revisit it based on operational experience and user feedback.

tuunit (approved):

PR #323, opened 2026-05-18
chore: increase SKE nodepool disk size

Matthiator:
## 📝 Summary

Increase the generated STACKIT SKE node pool root disk size from the provider default by making `volume_size` explicit.

- add `volume_size` to the generated SKE node pool input schema with a default of `30` GB
- set the generated example infrastructure node pool to `volume_size = 30`
- update the STACKIT Terraform README examples and module input docs

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [x] 🧱 Terraform module
- [x] 📝 Documentation
- [ ] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

No. According to the Terraform integration test, changing the node pool `volume_size` is planned as an in-place update.

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

Terraform integration test confirmed this is an in-place update.

## 🔗 Related Issues / Tickets
N/A

## ✅ Checklist

During the integration tests we saw node-level disk pressure/image garbage collection symptoms. This is separate from service PVC sizing: Prometheus and other stateful services can still use PVCs, while kubelet/container runtime images and writable node data use the SKE node pool root volume.

If `volume_size` is not set, the provider default is 20 GB. This PR uses 30 GB as a conservative first step that gives the node more headroom without making the default too opinionated. The right value still depends on workload, image churn, rollout patterns, and logging volume, and users can override it per node pool when they need more space.

Validation included `go test ./...` from `go-binary` and `terraform fmt -check go-binary/templates/embedded/managed-service-catalog/terraform/providers/stackit/modules/ske-cluster/variables.tf`.

### DIFF
--- a/go-binary/templates/embedded/customer-service-catalog/terraform/providers/stackit/example/infrastructure/env.auto.tfvars.tplt
+++ b/go-binary/templates/embedded/customer-service-catalog/terraform/providers/stackit/example/infrastructure/env.auto.tfvars.tplt
@@ -29,6 +29,7 @@ node_pools = [
   {
     availability_zones = ["eu01-2"]
     machine_type       = "c2i.8"
+    volume_size        = 30
     maximum            = 4
     minimum            = 2
     name               = "pool-infra"
--- a/go-binary/templates/embedded/customer-service-catalog/terraform/providers/stackit/example/infrastructure/variables.tf.tplt
+++ b/go-binary/templates/embedded/customer-service-catalog/terraform/providers/stackit/example/infrastructure/variables.tf.tplt
@@ -92,6 +92,7 @@ List of node pools. Each element must be an object with:
 - name               = string
 - machine_type       = string
 - os_version_min     = optional(string)
+- volume_size        = optional(number, 30)
 - minimum            = number
 - maximum            = number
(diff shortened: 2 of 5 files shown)


go-binary/templates/embedded/customer-service-catalog/terraform/providers/stackit/example/infrastructure/env.auto.tfvars.tplt
```diff
@@ -29,6 +29,7 @@ node_pools = [
   {
     availability_zones = ["eu01-2"]
     machine_type       = "c2i.8"
+    volume_size        = 40
```
tuunit: why 40?
20 is the current default, right?
wouldn't 30 suffice?

Matthiator: Yes, exactly, 20 is the default. We can start with 30. Users will then need to adjust the value themselves later according to their needs.

tuunit (approved):

PR #347, opened 2026-05-28
feat: refactor stackit edge cloud provisioning

Matthiator:
## 📝 Summary

  
Refactors STACKIT Edge Cloud provisioning so Edge Cloud can be handled as a separate, opinionated provisioning path next to SKE.

This PR:
- Adds separate Terraform modules for Edge Cloud instance creation, Edge image upload, and Edge host provisioning.
- Reworks the generated STACKIT Terraform example for the new Edge Cloud flow.
- Splits STACKIT provider docs into Terraform bootstrap, SKE provisioning, and Edge Cloud provisioning pages.
- Documents the current single-node Edge Cloud ingress approach with STACKIT public IP, private VM IP, MetalLB, and external-dns.
- Updates generated Helm/Terraform templates and tests for the new Edge Cloud shape.
- Fixes Longhorn dashboard exposure below the `/longhorn` Traefik subpath by rendering the required redirect and path rewrite middlewares.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [x] 📦 Helm chart
- [x] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

The previous Edge Cloud Terraform input/output shape is replaced. Legacy single-host Edge inputs are intentionally removed in favor of the new Edge instance, image, and hosts structure.

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested
  
Manual validation so far:
- Built the kubara binary locally and generated Terraform/Helm artifacts for an Edge Cloud test repository.
- Applied the Edge Cloud Terraform flow far enough to provision the Edge instance, uploaded image, and VM host.
- Bootstrapped Argo CD and deployed the managed Helm stack on the Edge Cloud cluster.
- Validated Longhorn dashboard routing behind the Traefik `/longhorn` subpath.
- Ran `make docs-build` successfully.
- Ran `make test` successfully.

Still pending before marking this PR ready:
- Full CI result.

## ✅ Checklist
  
This PR is intentionally opened as draft while the Edge Cloud flow is still being validated end-to-end.


docs/content/1_getting_started/providers/stackit_provisioning_edgecloud.md
```diff
@@ -0,0 +1,376 @@
...
+- UI is often the fastest manual path for image and cluster creation.
+- CLI/API with manifests is better for reproducibility and change history (for example in Git).
+
+For general STEC background, see the official [Edge Cloud overview](https://docs.stackit.cloud/products/runtime/edge-cloud/), [Authentication](https://docs.stackit.cloud/products/runtime/edge-cloud/getting-started/authentication/), and [Using the API](https://docs.stackit.cloud/products/runtime/edge-cloud/tutorials/using-the-api/) guides.
+
+Choose the operating model that fits your team. kubara examples use manifest-based flows for traceability, but this is not mandatory.
+
+## Terraform modules for the kubara Edge example
+
+The generated Edge Terraform example is intentionally create-only:
```
fabian-schmitt: i think this reads a bit weird, might use "is intentionally creating only the following resources:" ?

docs/content/1_getting_started/providers/stackit_provisioning_edgecloud.md
```diff
@@ -0,0 +1,376 @@
...
+
+## Single-node bootstrap and network planning
+
+The generated `edge_hosts.nodes` defaults to one control plane node as a bootstrap example.
+This is intentionally simple for initial rollout and troubleshooting.
+
+For production and multi-node topologies, design networking explicitly for your environment:
+
+* where hosts run (VMs, bare metal, mixed, multi-cloud)
+* ingress entry strategy (single public node, MetalLB VIP, external NLB, etc.)
```
fabian-schmitt: maybe rather call it "reachability (strategy)"

docs/content/1_getting_started/providers/stackit_provisioning_edgecloud.md
```diff
@@ -0,0 +1,376 @@
...
+
+```hcl
+edge_image = {
+  local_file_path = "edge-artifacts/openstack-amd64.raw"
+  # ...
+}
+```
+
+6. Configure `edge_hosts.nodes`.
+7. Run full apply. Terraform uploads the image and creates the VM/network-based hosts from that uploaded image:
```
fabian-schmitt: What is meant by "the VM/network-based hosts"? 
Can this be replaced by just "Host", i found this to be confusing

docs/content/1_getting_started/providers/stackit_terraform_bootstrap.md
```diff
@@ -0,0 +1,93 @@
+# STACKIT Terraform Bootstrap
+
+We recommend using Terraform to provision your Kubernetes and additional infrastructure like a Secret Manager instance.
```
fabian-schmitt: Do we really recommend? I am not sure about this, since this sentence doesn't sound platform agnostic to me

fabian-schmitt (changes requested):
see comments

docs/content/1_getting_started/providers/stackit_provisioning_edgecloud.md
Matthiator: Agreed. I'll rephrase this to say that the generated example intentionally creates only the listed resources, which reads clearer than `create-only`.

docs/content/1_getting_started/providers/stackit_provisioning_edgecloud.md
Matthiator: Agreed. I'll use `external reachability strategy` here and spell out `Network Load Balancer` instead of using the `NLB` abbreviation.

docs/content/1_getting_started/providers/stackit_provisioning_edgecloud.md
Matthiator: Good point. The intent is STACKIT VM hosts plus their network resources. I'll replace this wording with that more explicit phrasing.

docs/content/1_getting_started/providers/stackit_terraform_bootstrap.md
Matthiator: Good catch. The intent is STACKIT-specific Terraform support, not a platform-agnostic Terraform recommendation. I'll rephrase this accordingly.

src/cmd/generate_test.go
tuunit, line 1: do we really need a unit test for templated files? unit tests should be about code testing not template tests, right?

src/internal/catalog/built-in/customer-service-catalog/terraform/providers/stackit/example/infrastructure/outputs.tf.tplt
```diff
@@ -58,18 +58,44 @@ output "kubeconfig_raw" {
...
+output "edge_status" {
+  description = "Status of the Edge Cloud instance"
+  value       = module.edge_instance.status
+}
+
+output "edge_kubeconfig_raw" {
+  description = "Raw Edge Cloud kubeconfig (short-lived, sensitive)"
+  value       = module.edge_instance.kubeconfig_raw
+  sensitive   = true
+}
```
tuunit: what would be a "none raw" kubeconfig? i think removing the _raw suffix would be a good idea as I don't think this additional word has any meaning in this context

src/internal/catalog/built-in/customer-service-catalog/terraform/providers/stackit/example/infrastructure/variables.tf.tplt
```diff
@@ -165,31 +165,89 @@ variable "expiration" {
...
+
+  validation {
+    condition     = trimspace(var.edge_instance.display_name) != ""
+    error_message = "edge_instance.display_name must be set."
+  }
+
+  validation {
+    condition     = can(regex("^[a-z0-9][a-z0-9-]{2,6}[a-z0-9]$", trimspace(var.edge_instance.display_name)))
+    error_message = "edge_instance.display_name must be a valid hostname label with 4 to 8 characters."
+  }
```
tuunit: bit of an overkill in my opinion as the api would most likely tell you the same anyways, right? but as we already have it now. it should be fine to keep it here

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/ske-cluster/README.md
```diff
@@ -28,7 +28,7 @@ module "ske_cluster" {
     },
     {
       name               = "np-backend"
-      machine_type       = "c1.2"
+      machine_type       = "c2i.8"
```
tuunit, line 31: wouldn't c2i.4 suffice?

docs/mkdocs.yml
```diff
@@ -14,7 +14,12 @@ nav:
       - Prerequisites: 1_getting_started/prerequisites.md
       - Bootstrapping: 1_getting_started/bootstrapping.md
       - Providers:
-          - STACKIT: 1_getting_started/providers/stackit.md
+          - STACKIT:
+              - Overview: 1_getting_started/providers/stackit.md
+              - Terraform Bootstrap: 1_getting_started/providers/stackit_terraform_bootstrap.md
+              - Provisioning Infrastructure:
+                  - SKE: 1_getting_started/providers/stackit_provisioning_ske.md
+                  - Edge Cloud (Beta): 1_getting_started/providers/stackit_provisioning_edgecloud.md
```
tuunit, line 20: I'm questioning this heavy nesting of sub sub sub pages 🤔 
We should find a better way that is less convoluted before merging

tuunit: Lets wait for the team decision on the name change for providers and I'll restructure accordingly

tuunit (changes requested):

src/cmd/generate_test.go
Matthiator: Agreed. Folded the edge case into the table-driven `TestGenerateCmd` and replaced the assertions on rendered Terraform with generator-behavior checks (expected artifacts produced at the right path, provider folders don't leak) via `os.Stat`/`os.ReadDir`, like the other cases. Also tightened `AGENTS.md` so this expectation explicitly covers Terraform templates, not just Helm. (67d939d3)

src/internal/catalog/built-in/customer-service-catalog/terraform/providers/stackit/example/infrastructure/outputs.tf.tplt
Matthiator: Good catch — `_raw` was our own convention, not an upstream name (the provider attributes are `kube_config`/`kubeconfig`). Renamed `kubeconfig_raw` → `kubeconfig` and `edge_kubeconfig_raw` → `edge_kubeconfig` consistently across the SKE + Edge modules, the example template, docs and module READMEs, and also dropped the now-meaningless "Raw" from the descriptions. It isn't breaking: a Terraform output rename just updates state on the next apply (no resource recreation) and nothing in the Go/bootstrap code consumes the name. (1a445e3a)

src/internal/catalog/built-in/customer-service-catalog/terraform/providers/stackit/example/infrastructure/variables.tf.tplt
Matthiator: Removed the regex validation (kept only the non-empty check) — the STACKIT API is the source of truth for the name format. While at it, the edge instance name is now derived from the cluster name + a random suffix and trimmed to the 8-character limit (same pattern as the SKE cluster name) instead of being a free-form input, so `display_name` is no longer set in tfvars. (e4484e85)

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/ske-cluster/README.md
Matthiator: We've only validated the stack on `c2i.8` (8 vCPU / 16 GB) so far, so I'd like to keep the documented example at the tested flavor for now. `c2i.4` would likely suffice for multi-node SKE, but I'd rather not change the documented value before verifying it — and as a related data point, on the single-node Edge Cloud setup 8 GB was already too tight in my testing (not all pods came up). Proposal: keep `c2i.8` here and revisit `c2i.4` in a follow-up once tested.

docs/mkdocs.yml
Matthiator: Agreed — keeping the nesting as-is for now and not blocking this PR on it. I'll do a follow-up PR to flatten the provider docs structure once the T Cloud Public / Edge Cloud docs are merged and the provider naming decision is settled.

src/cmd/generate_test.go
```diff
@@ -206,6 +207,51 @@ func TestGenerateCmd(t *testing.T) {
...
 			},
 		},
+		{
+			name:    "successful edge terraform file generation",
+			flags:   []string{"--terraform"},
+			wantErr: false,
+			cluster: &config.Cluster{
+				Name:    "edge-cluster",
+				Stage:   "dev",
+				Type:    "controlplane",
```
tuunit:
```suggestion
				Type:    "hub",
```

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/edge-hosts/main.tf
```diff
@@ -0,0 +1,95 @@
...
+
+  security_group_ids = [stackit_security_group.this.security_group_id]
+}
+
+resource "stackit_server" "this" {
+  # Create one VM per configured Edge host.
+  for_each = local.nodes_by_name
+
+  project_id   = var.project_id
+  name         = each.value.name
```
tuunit: why isn't this formatted / prefixed like the other resources?

e.g.:
`name = "${var.name}-${each.key}-server"`

tuunit: Personally I think it is a bit unnecessary to suffix using -server or -nic or -volume. As the resources cannot be mixed and aren't listed together.

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/edge-hosts/main.tf
```diff
@@ -0,0 +1,95 @@
...
+    max = tonumber(each.value)
+  }
+}
+
+resource "stackit_volume" "this" {
+  # Create one boot volume per configured Edge host.
+  for_each = local.nodes_by_name
+
+  project_id        = var.project_id
+  name              = "${var.name}-${each.key}-volume"
```
tuunit:
```suggestion
  name              = "${var.name}-${each.key}"
```

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/edge-hosts/main.tf
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+resource "stackit_network_interface" "this" {
+  # Create one NIC per configured Edge host.
+  for_each = local.nodes_by_name
+
+  project_id = var.project_id
+  network_id = stackit_network.this.network_id
+  name       = "${var.name}-${each.key}-nic"
```
tuunit:
```suggestion
  name       = "${var.name}-${each.key}"
```

tuunit (changes requested):

src/cmd/generate_test.go
Matthiator: Good catch — `controlplane` isn't a valid cluster type (the schema enum is hub/spoke), so the test was using an invalid value. Applied in a8677214.

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/edge-hosts/main.tf
Matthiator: Applied in 99d09673 — dropped the `-volume` suffix.

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/edge-hosts/main.tf
Matthiator: Applied in 99d09673 — dropped the `-nic` suffix.

src/internal/catalog/built-in/managed-service-catalog/terraform/providers/stackit/modules/edge-hosts/main.tf
Matthiator: Agreed, and good point on dropping the suffixes. Aligned all three per-node resources (volume, network interface, server) on `name = "${var.name}-${each.key}"` — the server previously used the bare node name with no prefix. The network and security group keep their dedicated `network_name`/`security_group_name` variables since those are user-provided. Done in 99d09673.

tuunit (approved):
LGTM

PR #362, opened 2026-06-01
fix: refresh helm repo cache during bootstrap

Matthiator:
## 📝 Summary

Make `kubara bootstrap` robust against stale Helm repository caches that can lead to "can't get a valid version for subchart" errors while preparing bootstrap chart dependencies.

The PR now keeps the fix focused on refreshing Helm repository metadata and preserving dependency pinning:

- New `helm.UpdateAllRepositories` runs `helm repo update` without arguments after adding the enabled bootstrap chart repos, so Helm's shared repository cache is current even when `Chart.yaml` references subcharts by repository URL rather than by alias.
- Bootstrap dependency preparation now uses `helm dependency build`, which respects an existing `Chart.lock` and only mirrors `helm dependency update` behavior when no lock file exists.
- Bootstrap chart repo/dependency handling is now gated by an explicit `Enabled` flag instead of `EnsureCRD`, so CRD handling stays separate from whether a chart participates in the bootstrap flow.

This avoids the previous manual workaround of running `helm repo update` before bootstrap while not deleting `Chart.lock` or undermining reproducible dependency resolution.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [ ] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

No breaking change expected. The bootstrap flow performs one global Helm repository refresh for the enabled bootstrap chart repos and uses `helm dependency build` to preserve locked dependency versions when a `Chart.lock` is present.

## 🧪 Testing
- [ ] CI passed
- [ ] Unit tested

- `cd src && make test` green locally
- `cd src && make build` green locally
- Manually reproduced the original error during a `kubara bootstrap` against a T Cloud Public CCE cluster and confirmed the refreshed repo cache flow recovers without a manual `helm repo update`

## 🔗 Related Issues / Tickets

n/a — surfaced while bootstrapping a new T Cloud Public cluster.

## ✅ Checklist

No docs change — this removes a bootstrap failure mode without changing the documented success path.

Original error during bootstrap that motivated this PR:

```text
FTL Error running program error="update helm repositories: update helm chart dependencies for \"kube-prometheus-stack\":
helm dependency update failed for .../managed-service-catalog/helm/kube-prometheus-stack: exit status 1
Stderr: Error: can't get a valid version for 1 subchart(s): \"kube-prometheus-stack\"
(repository \"https://prometheus-community.github.io/helm-charts\", version \"86.1.0\").
Make sure a matching chart version exists in the repo, or change the version constraint in Chart.yaml"
```

The relevant bootstrap flow now is:

```go
if err := helm.UpdateAllRepositories(ctx); err != nil {
    return fmt.Errorf("refresh helm repository index: %w", err)
}
```

and dependency preparation now uses:

```go
if err := helm.BuildDependencies(ctx, dep); err != nil {
    return fmt.Errorf("build helm chart dependencies for %q: %w", chart.Name, err)
}
```


src/internal/cmd/bootstrap/bootstrap.go
```diff
@@ -205,9 +206,22 @@ func addHelmRepositories(ctx context.Context, charts []BootstrapChart) error {
...
+			added = true
 		}
 	}
 
+	// Refresh the global repository index too. Per-alias updates can leave the
+	// shared cache out of date for charts whose Chart.yaml references the
+	// repository by URL rather than by the alias added above, which triggers
+	// "can't get a valid version for subchart" errors on the next
+	// helm dependency update.
+	if added {
```
tuunit: i think this logic is unnecessary we should just run helm repo update every time

src/internal/cmd/bootstrap/bootstrap.go
```diff
@@ -194,6 +194,7 @@ func ensureNamespaces(ctx context.Context, client *k8s.Client, opts *Options, ch
 func addHelmRepositories(ctx context.Context, charts []BootstrapChart) error {
 	log.Info().Msg("Adding helm repositories")
 
+	added := false
 	for _, chart := range charts {
 		if chart.EnsureCRD {
```
tuunit: not a change that has been introduced by this PR but why do we add charts and repos only if ensure crds is set? that doesn't make sense

src/internal/cmd/bootstrap/bootstrap.go
```diff
@@ -217,6 +231,14 @@ func updateHelmDependencies(ctx context.Context, opts *Options, charts []Bootstr
 
 	for _, chart := range charts {
 		if chart.EnsureCRD {
```
tuunit: same question here? why only for charts with ensure CRD?

src/internal/cmd/bootstrap/bootstrap.go
```diff
@@ -217,6 +231,14 @@ func updateHelmDependencies(ctx context.Context, opts *Options, charts []Bootstr
...
+			// Discard any previously downloaded subchart archives and the lock
+			// file so the next dependency update resolves against the freshly
+			// refreshed repository index. Otherwise a stale Chart.lock can pin
+			// kubara to a subchart version that no longer matches the cache.
```
tuunit: I don't agree with this statement 🤔 This would undermine secure version pinning

src/internal/cmd/bootstrap/bootstrap.go
Matthiator: Agree. I removed the `added` guard and now run the global `helm repo update` unconditionally after adding the enabled bootstrap chart repos. I also removed the redundant per-repo update calls, so we only refresh once globally.

src/internal/cmd/bootstrap/bootstrap.go
Matthiator: Good point. I introduced an explicit `Enabled` flag for bootstrap charts and changed repository setup to use that instead of `EnsureCRD`. `EnsureCRD` now only controls CRD application, while repo setup follows whether the chart participates in the bootstrap flow.

src/internal/cmd/bootstrap/bootstrap.go
Matthiator: Same fix here: dependency handling now uses `chart.Enabled` instead of `chart.EnsureCRD`. That keeps dependency preparation separate from the CRD-specific behavior.

src/internal/cmd/bootstrap/bootstrap.go
Matthiator: Agreed. I removed the cleanup entirely so we no longer delete `Chart.lock` or `charts/`. I also switched bootstrap dependency preparation from `helm dependency update` to `helm dependency build`, which respects an existing `Chart.lock` and only mirrors update behavior when no lock file exists. That should preserve secure version pinning while still fixing the stale repo cache issue via the global repo refresh.

tuunit (approved):

PR #367, opened 2026-06-03
fix: include additional values during bootstrap

Matthiator:
## 📝 Summary

Direct `kubara bootstrap` now loads chart-local `additional-values.yaml` files when they exist. This aligns the bootstrap path with the generated Argo CD ApplicationSets, which already reference both `values.yaml` and `additional-values.yaml`.

Previously, bootstrap only passed `values.yaml` to Helm. As a result, overrides in `customer-service-catalog/helm/<cluster>/argo-cd/additional-values.yaml` were ignored unless Argo CD managed itself.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

Ran:

```bash
cd src && go test ./internal/cmd/bootstrap
make docs-build
```

## 🔗 Related Issues / Tickets

Related to the generated values overwrite behavior tracked separately.

## ✅ Checklist

This PR intentionally only addresses direct bootstrap. It does not change `kubara generate --helm` overwrite behavior for generated `values.yaml` files.

### DIFF
--- a/docs/content/1_getting_started/bootstrapping.md
+++ b/docs/content/1_getting_started/bootstrapping.md
@@ -195,7 +195,7 @@ Every generated app supports:
 * optional `additional-values.yaml` for overrides/extra values
 
 kubara's generated ApplicationSet already references both files and ignores missing files, so you can add `additional-values.yaml` only when needed.
-During direct bootstrap, kubara also loads chart-local `additional-values.yaml` files when they exist.
+During bootstrap kubara uses `additional-values.yaml` files when they exist as well.
 
 Merge behavior reminder:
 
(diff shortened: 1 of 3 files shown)


docs/content/1_getting_started/bootstrapping.md
```diff
@@ -195,6 +195,7 @@ Every generated app supports:
 * optional `additional-values.yaml` for overrides/extra values
 
 kubara's generated ApplicationSet already references both files and ignores missing files, so you can add `additional-values.yaml` only when needed.
+During direct bootstrap, kubara also loads chart-local `additional-values.yaml` files when they exist.
```
tuunit: I think this is unnecessary to mention or at least if we keep it rephrase it to:


```suggestion
During bootstrap kubara uses `additional-values.yaml` files when they exist as well.
```

tuunit (changes requested):
Rest looks good to me

PR #372, opened 2026-06-05
feat!: catalog versioning, packaging and caching

tuunit:
## 📝 Summary
This PR introduces new kubara catalog management functionality by introducing three new CLI subcommands for listing, packaging, and unpackaging catalogs, with support for OCI (Open Container Initiative) references. The changes also introduces and enforces versioning for catalogs for better platform lifecycle management.

As this is just the first step of the catalog management, the documentation was intentionally not updated (besides the auto-generated command page) and we will wait until the next PR is ready for introducing pulling, pushing and login to remote registries before fully reworking the catalog pages.

**Breaking change:**
This PR introduces and enforces the catalog version to be set. Therefore if users already have created their own custom catalogs they will have to manually add a version field to their `Catalog.yaml`:

```yaml
apiVersion: kubara.io/v1alpha1
kind: Catalog
metadata:
  name: your-catalog
spec:
  version: 0.1.0
```

## 🧩 Type of change
- [X] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [X] 📝 Documentation
- [X] 🧪 Test or CI change
- [X] ♻️ Refactor / cleanup

- [X] Yes, this change breaks existing functionality (explain in summary)

## 🧪 Testing
- [X] CI passed
- [X] Unit tested
  
## ✅ Checklist

### DIFF
--- a/src/internal/catalog/cache.go
+++ b/src/internal/catalog/cache.go
@@ -180,7 +180,9 @@ func copyFile(sourcePath, targetPath string, mode fs.FileMode) error {
 	if err != nil {
 		return fmt.Errorf("open cached catalog file %q: %w", sourcePath, err)
 	}
-	defer sourceFile.Close()
+	defer func() {
+		_ = sourceFile.Close()
+	}()
 
 	targetFile, err := os.OpenFile(targetPath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, mode)
 	if err != nil {
--- a/src/internal/catalog/cache.go
+++ b/src/internal/catalog/cache.go
@@ -1,0 +1,220 @@
+package catalog
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/kubara-io/kubara/internal/utils"
+)
+
+type CachedCatalogEntry struct {
+	Reference      string
+	CatalogName    string
+	CatalogVersion string
+	ManifestDigest string
+}
+
+type UnpackageResult struct {
+	Artifact   CachedArtifact
+	OutputPath string
+}
+
+func ListCachedCatalogs() ([]CachedCatalogEntry, error) {
+	cacheRoot, err := defaultCatalogCacheRoot()
+	if err != nil {
+		return nil, err
+	}
+
+	entries := make(map[string]CachedCatalogEntry)
+	referencedDigests := make(map[string]struct{})
+
+	if err := collectReferenceEntries(filepath.Join(cacheRoot, "refs"), entries, referencedDigests); err != nil {
+		return nil, err
+	}
+
+	result := make([]CachedCatalogEntry, 0, len(entries))
+	for _, entry := range entries {
+		result = append(result, entry)
+	}
+
+	// Sort by name, then version and then source reference
+	slices.SortFunc(result, func(a, b CachedCatalogEntry) int {
+		if cmp := strings.Compare(a.CatalogName, b.CatalogName); cmp != 0 {
+			return cmp
+		}
+		if cmp := strings.Compare(a.CatalogVersion, b.CatalogVersion); cmp != 0 {
+			return cmp
+		}
+		return strings.Compare(a.Reference, b.Reference)
+	})
+
+	return result, nil
+}
+
+func UnpackageCatalog(options UnpackageOptions) (UnpackageResult, error) {
+	artifact, err := GetCachedArtifact(options.Reference)
+	if err != nil {
+		return UnpackageResult{}, err
+	}
+
+	outputPath := strings.TrimSpace(options.OutputPath)
+	if outputPath == "" {
+		outputPath, err = utils.GetFullPath(artifact.CatalogName, options.WorkDir)
+		if err != nil {
+			return UnpackageResult{}, fmt.Errorf("resolve default unpack directory: %w", err)
+		}
+	}
+
+	if _, err := os.Stat(outputPath); err == nil {
+		return UnpackageResult{}, fmt.Errorf("catalog output directory %q already exists", outputPath)
+	} else if !os.IsNotExist(err) {
+		return UnpackageResult{}, fmt.Errorf("stat catalog output directory %q: %w", outputPath, err)
+	}
+
+	if err := copyDirectory(artifactRootPath(artifact), outputPath); err != nil {
+		return UnpackageResult{}, err
+	}
+
+	return UnpackageResult{
+		Artifact:   artifact,
+		OutputPath: outputPath,
+	}, nil
+}
+
+func collectReferenceEntries(root string, entries map[string]CachedCatalogEntry, referencedDigests map[string]struct{}) error {
+	return filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			if os.IsNotExist(err) {
+				return nil
+			}
+			return fmt.Errorf("walk cached catalog references: %w", err)
+		}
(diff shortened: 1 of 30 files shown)


src/internal/catalog/cache.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	defer sourceFile.Close()
+
+	targetFile, err := os.OpenFile(targetPath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, mode)
+	if err != nil {
+		return fmt.Errorf("create catalog output file %q: %w", targetPath, err)
+	}
+
+	if _, err := io.Copy(targetFile, sourceFile); err != nil {
+		targetFile.Close()
```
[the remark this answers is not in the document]
tuunit: Resolved

src/internal/catalog/source_resolver.go
```diff
@@ -0,0 +1,151 @@
...
+func tryLoadCatalogManifest(root string) (*CatalogManifest, error) {
+	path := filepath.Join(root, "Catalog.yaml")
+	if _, err := os.Stat(path); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("stat %q: %w", path, err)
+	}
+
+	manifest, err := LoadCatalogManifest(root)
```
Matthiator, line 136: Preserve compatibility for catalogs without `spec.version`. Catalogs produced by previous Kubara versions contain a `Catalog.yaml` without a version. Loading the manifest strictly here makes those catalogs fail in `catalog add`, `generate`, and `schema`. Please provide backward-compatible loading or a migration path; otherwise this needs to be explicitly treated and documented as a breaking change in this PR.

src/internal/catalog/source_resolver.go
```diff
@@ -0,0 +1,151 @@
...
+	return &manifest, nil
+}
+
+func resolveServicesPath(catalogPath string) (string, error) {
+	cleaned, err := requireDirectory(catalogPath, "catalog path")
+	if err != nil {
+		return "", err
+	}
+
+	servicesDir := filepath.Join(cleaned, string(ServicesDirectory))
```
Matthiator, line 149: Keep the documented direct services-directory behavior. The previous resolver fell back to `catalogPath` when no nested `services/` directory existed, and the current docs/ADR explicitly allow `--catalog` to point directly at a services directory. This now looks for `services/services` and fails. Please restore the fallback or explicitly migrate this supported contract.

src/internal/catalog/cache_store.go
```diff
@@ -0,0 +1,232 @@
...
+		Tag:            ref.Tag,
+		UpdatedAt:      time.Now().UTC().Format(time.RFC3339),
+	})
+}
+
+func writeReferenceFile(path string, ref cachedReference) error {
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return fmt.Errorf("create cached catalog reference directory: %w", err)
+	}
+	return writeJSONFile(path, "cached catalog reference", ref)
```
Matthiator, line 224: Avoid orphaning artifacts when a version reference is updated. Repackaging changed content under the same version finalizes a new digest and overwrites this tag reference, but the old artifact remains under `artifacts/`. Since `ListCachedCatalogs` only walks `refs/`, that artifact becomes invisible and repeated packaging grows the cache indefinitely. Either make version references immutable or garbage-collect digests that are no longer referenced.

src/internal/catalog/package.go
```diff
@@ -0,0 +1,158 @@
...
+	contentsDir := filepath.Join(tempArtifactDir, "contents")
+	rootDir, err := extractCatalogContents(ctx, layoutStore, manifestDescriptor, contentsDir)
+	if err != nil {
+		return CachedArtifact{}, err
+	}
+
+	artifact.ManifestDigest = manifestDescriptor.Digest.String()
+	artifact.RootDirectory = filepath.Base(rootDir)
+
+	if err := finalizeCachedArtifact(tempArtifactDir, artifact); err != nil {
```
Matthiator, line 116: Remove the staging copy before finalizing. `finalizeCachedArtifact` renames the entire temporary directory, which still contains the complete `staging/` tree in addition to the OCI layout and extracted `contents/`. Every cached artifact therefore retains an unnecessary additional full copy of the catalog.

src/internal/catalog/cache_store.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	return nil
+}
+
+func writeJSONFile(path, label string, value any) error {
+	raw, err := json.MarshalIndent(value, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshal %s: %w", label, err)
+	}
+	if err := os.WriteFile(path, raw, 0o600); err != nil {
```
Matthiator: Reference updates should be atomic. `os.WriteFile` truncates an existing reference before writing the replacement, so interruption or a write failure can leave invalid JSON and break subsequent list/load operations. Write to a temporary file in the same directory, close/sync it, and rename it over the reference.

src/internal/catalog/cache.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+	}
+
+	if _, err := os.Stat(outputPath); err == nil {
+		return UnpackageResult{}, fmt.Errorf("catalog output directory %q already exists", outputPath)
+	} else if !os.IsNotExist(err) {
+		return UnpackageResult{}, fmt.Errorf("stat catalog output directory %q: %w", outputPath, err)
+	}
+
+	if err := copyDirectory(artifactRootPath(artifact), outputPath); err != nil {
```
Matthiator, line 80: Clean up a partially materialized catalog when copying fails. `copyDirectory` may already have created directories and files before returning an error; the next invocation then always fails because the output path exists. Copying to a temporary sibling followed by rename, or removing the newly created output on failure, would keep retries safe.

Matthiator (changes requested):
Requesting changes because this introduces two undocumented upgrade breaks and several cache lifecycle/integrity problems. Enforcing catalog versions is intentional, but catalogs created by previous Kubara versions have no `spec.version` and there is no migration path. The broader catalog documentation can follow later, but the breaking behavior and required migration need to be handled in this PR. Please also add focused tests for OCI reference parsing and cache replacement/failure scenarios. The currently failing `go-static-analysis` check still reports three unchecked `Close()` errors.

src/internal/catalog/source_resolver.go
tuunit: Explicitly marked the PR as a breaking change with `feat!:` and will document it in the PR description as well

src/internal/catalog/source_resolver.go
tuunit: I think the previous design was just plainly a stupid wrong idea. Why would we want to allow services directories directly. This new design/api contract is the better approach moving forward. I'll change the docs accordingly and not implement a fallback behaviour. As we are still in v0 it is acceptable to break a couple small things, at least in my humble opinion.

src/internal/catalog/cache_store.go
tuunit: Good point but something for future improvements. Maybe a prune command or something of the sort could be introduced in future versions

src/internal/catalog/cache.go
tuunit: nah i think that is fine. i really don't want to deal with directory cleanup / deletion as this could introduce unnecessary risk of users being in a system path and we could potentially delete something important. in that case the user should rather be the one to clean it up.

Instead we already have a warning/error if the outputPath already exists and therefore copyDirectory will not be executed anyways without the user cleaning up the directory beforehand

Matthiator (approved):
lgtm

PR #370, opened 2026-06-04
feat: add T Cloud Public CCE provider support

Matthiator:
## 📝 Summary

Adds T Cloud Public CCE provider support, including generated Terraform stacks for backend bootstrap, cluster infrastructure, OpenBao configuration, provider-specific Helm values, and setup documentation.

Main pieces:
- T Cloud Public provider docs and setup flow
- Terraform modules for OBS buckets, IAM agencies, DNS, network, keypair, KMS, CCE, storage classes, and OpenBao Helm
- Generated infrastructure and OpenBao Terraform layers
- Provider-specific ExternalDNS, External Secrets, Traefik, OpenBao, and Velero wiring
- Namespace-isolated OpenBao secret access for the T Cloud Public path
- Shared Terraform override guidance for generated `env.auto.tfvars`

This draft currently includes PR #366 in its diff because this branch was built on top of that Velero work. Please merge #366 first; after that this branch should be rebased onto `main` so this PR only shows the T Cloud Public delta.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [x] 📦 Helm chart
- [x] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

## ⚠️ Is this a breaking change?
- [ ] Yes, this change breaks existing functionality (explain in summary)

No intended breaking changes. T Cloud Public is added as a new provider path.

## 🧪 Testing
- [ ] CI passed
- [ ] Manually tested (local/dev cluster)
- [x] Unit tested
- [ ] Not tested (explain why below)

Ran locally:
- `make test`
- `make build-binary`
- `make docs-validate`
- `make docs-build`
- `terraform fmt -check -recursive src/internal/catalog/built-in/managed-service-catalog/terraform/providers/t-cloud-public/modules/network`
- `terraform init -backend=false` and `terraform validate` for the T Cloud Public network module
- `git diff --check`

Cluster-level validation should be confirmed before marking this PR ready for review.

## 🔗 Related Issues / Tickets
Depends on #366

## ✅ Checklist
- [x] Code compiles and passes all tests
- [x] Linting and style checks pass
- [x] Comments added for complex logic
- [x] Documentation updated (if applicable)

## 📎 Additional Context (optional)
Open as draft until PR #366 is merged and this branch has been rebased onto the updated `main`.

PR #391, opened 2026-06-11
refactor: make external-dns values provider-aware

Matthiator:
## 📝 Summary

Prepares the first small split from #370 by refactoring the built-in ExternalDNS Helm values for provider-specific configuration.

The existing Stackit-specific ExternalDNS settings are now rendered conditionally from `terraform.provider`, while provider-independent values remain shared. This establishes the supported pattern for adding further provider-specific ExternalDNS configuration without duplicating complete Helm values files.

Additional changes made while preparing this split and addressing review feedback:
- Treat provider-specific template directories only under `terraform/providers/<provider>/`; Helm provider behavior stays in shared values templates or charts.
- Introduce a dedicated `TerraformProvider` type and centralize supported Terraform provider validation.
- Introduce `terraform.provider: none` as the default to support clusters without Terraform-managed infrastructure.
- Skip Terraform template generation for clusters without Terraform config during default `kubara generate`.
- Keep `kubara generate --terraform` strict and require a supported Terraform provider.
- Avoid rendering a provider-derived Velero `VolumeSnapshotClass` when Terraform is disabled.
- Handle existing clusters without Terraform during `kubara init --overwrite`.
- Document the Terraform-only provider template layout and the catalog boundary for app-specific behavior.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [x] 📦 Helm chart
- [ ] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

No intended breaking changes to supported behavior. Existing Stackit ExternalDNS values are preserved when `terraform.provider: stackit` is configured.

Clusters without Terraform config, or with `terraform.provider: none`, can now use the default `kubara generate` flow without forcing `--helm`.

## 🧪 Testing
- [x] CI passed
- [x] Unit tested

Ran locally:
- `cd src && make test`
- `cd src && go test ./internal/workflow ./cmd ./internal/config ./internal/render -count=1`
- `make docs-validate`
- `git diff --check`

## 🔗 Related Issues / Tickets
- First preparatory split from #370
- Follow-up to #167

## ✅ Checklist

This PR intentionally covers the ExternalDNS values preparation plus the provider handling needed to make `terraform.provider: none` a valid no-Terraform state.

It does not include T Cloud Public provider registration, OpenBao integration, Terraform modules, or the remaining provider-specific Helm changes from #370.

### DIFF
--- a/docs/content/1_getting_started/catalogs.md
+++ b/docs/content/1_getting_started/catalogs.md
@@ -194,9 +194,11 @@ During `kubara generate`, kubara renders templates and writes them into your rep
 - `example` path segments with the current cluster name
 
 Provider-specific template variants are supported only below `terraform/providers/<provider>/`, for example `customer-service-catalog/terraform/providers/stackit/example/infrastructure/main.tf.tplt`.
-When `terraform.provider` matches the provider name, kubara strips the `providers/<provider>/` path segment and uses that provider-specific file for the generated output path. If a common template and a provider-specific template resolve to the same output path, the provider-specific template replaces the common one; kubara does not deep-merge template files.
+When `terraform.provider` matches a supported provider name, kubara strips the `providers/<provider>/` path segment and uses that provider-specific file for the generated output path. If a common template and a provider-specific template resolve to the same output path, the provider-specific template replaces the common one; kubara does not deep-merge template files.
 
-Helm provider directories such as `helm/providers/<provider>/` are not supported. Keep provider-specific Helm behavior in a common values template or in the Helm chart itself.
+If a cluster has no Terraform block or uses `terraform.provider: none`, the default `kubara generate` command skips Terraform templates for that cluster. `kubara generate --terraform` still requires a supported Terraform provider.
+
+Provider directories below Helm paths are not treated as provider-specific overrides. Keep provider-specific Helm behavior in a common values template or in the Helm chart itself.
 
 ## Creating your own catalog
 
--- a/docs/content/1_getting_started/catalogs.md
+++ b/docs/content/1_getting_started/catalogs.md
@@ -193,6 +193,11 @@ During `kubara generate`, kubara renders templates and writes them into your rep
 - `customer-service-catalog/` with the configured customer overlay output path
 - `example` path segments with the current cluster name
 
+Provider-specific template variants are supported only below `terraform/providers/<provider>/`, for example `customer-service-catalog/terraform/providers/stackit/example/infrastructure/main.tf.tplt`.
+When `terraform.provider` matches the provider name, kubara strips the `providers/<provider>/` path segment and uses that provider-specific file for the generated output path. If a common template and a provider-specific template resolve to the same output path, the provider-specific template replaces the common one; kubara does not deep-merge template files.
+
+Helm provider directories such as `helm/providers/<provider>/` are not supported. Keep provider-specific Helm behavior in a common values template or in the Helm chart itself.
+
 ## Creating your own catalog
 
 If you want to understand how to create a custom catalog for your multi cluster platform have a look at [How to create your own catalog](../2_managing_your_platform/create_catalog.md)
--- a/src/cmd/generate_test.go
+++ b/src/cmd/generate_test.go
@@ -249,19 +249,7 @@ func TestGenerateCmd(t *testing.T) {
 				})
 
 				//dummy values
-				envPath := createTestEnv(t, tempDir, envconfig.EnvMap{
-					ProjectName:                 "project-name",
-					ProjectStage:                "project-stage",
-					DockerconfigBase64:          "DockerConfig",
-					ArgocdWizardAccountPassword: "wizardpassword",
-					ArgocdGitHttpsUrl:           "https://example.com",
-					ArgocdGitUsername:           "CoolCapybara",
-					ArgocdGitPatOrPassword:      "password",
-					ArgocdHelmRepoUrl:           "https://example.com",
-					ArgocdHelmRepoUsername:      "CoolCapybara",
-					ArgocdHelmRepoPassword:      "password",
-					DomainName:                  "example.com",
-				})
+				envPath := createDefaultGenerateTestEnv(t, tempDir)
 
 				// Add global flags
 				globalFlags := []string{
@@ -301,7 +289,7 @@ func TestGenerateCmd(t *testing.T) {
 	}
 }
 
-func TestGenerateCmd_MissingProviderUsesDefault(t *testing.T) {
+func TestGenerateCmd_MissingProviderDefaultsToNoneAndFailsForTerraform(t *testing.T) {
 	tempDir := t.TempDir()
 
 	configPath := createTestConfig(t, tempDir, config.Cluster{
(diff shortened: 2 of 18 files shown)


docs/content/1_getting_started/catalogs.md
```diff
@@ -193,6 +193,10 @@ During `kubara generate`, kubara renders templates and writes them into your rep
 - `customer-service-catalog/` with the configured customer overlay output path
 - `example` path segments with the current cluster name
 
+Provider-specific template variants can be placed below `providers/<provider>/` inside a `terraform` or `helm` template tree, for example `customer-service-catalog/helm/providers/stackit/example/external-dns/values.yaml.tplt`.
+When `terraform.provider` matches the provider name, kubara strips the `providers/<provider>/` path segment and uses that provider-specific file for the generated output path.
+If a common template and a provider-specific template resolve to the same output path, the provider-specific template replaces the common one; kubara does not deep-merge template files.
```
tuunit:
```suggestion
Infrastructure provider specific templates can be placed below `providers/<provider>/` inside a `terraform` or `helm` template tree, for example `customer-service-catalog/helm/providers/stackit/example/external-dns/values.yaml.tplt`.

- When `terraform.provider` matches the provider name, kubara strips the `providers/<provider>/` path segment and uses that provider-specific file for the generated output path.
- If a common template and a provider-specific template resolve to the same output path, the provider-specific template replaces the common one
- kubara does not deep-merge template files and instead does a full overwrite
```

src/internal/cmd/generate/generator.go
```diff
@@ -84,6 +84,17 @@ func resolveProvider(clusterBlock config.Cluster) (string, error) {
 	return provider, nil
 }
 
+func resolveProviderIfConfigured(clusterBlock config.Cluster) string {
+	if clusterBlock.Terraform == nil {
+		return ""
+	}
+	provider := strings.ToLower(strings.TrimSpace(clusterBlock.Terraform.Provider))
+	if provider == "" || provider == "<provider>" || !render.SupportedProviders[provider] {
```
tuunit:
```suggestion
	if !render.SupportedProviders[provider] {
```

shouldn't this be enough?

src/internal/cmd/generate/generator.go
```diff
@@ -84,6 +84,17 @@ func resolveProvider(clusterBlock config.Cluster) (string, error) {
 	return provider, nil
 }
 
+func resolveProviderIfConfigured(clusterBlock config.Cluster) string {
+	if clusterBlock.Terraform == nil {
+		return ""
+	}
+	provider := strings.ToLower(strings.TrimSpace(clusterBlock.Terraform.Provider))
```
tuunit: spaces are automatically trimmed by our yaml loader and don't we already validate enforce that the providers have to be in the support providers list while loading?

i mean is this conversion necessary at all?

src/internal/cmd/generate/generator.go
```diff
@@ -84,6 +84,17 @@ func resolveProvider(clusterBlock config.Cluster) (string, error) {
 	return provider, nil
 }
 
+func resolveProviderIfConfigured(clusterBlock config.Cluster) string {
+	if clusterBlock.Terraform == nil {
+		return ""
+	}
```
tuunit:
```suggestion
func resolveProviderIfConfigured(cluster config.Cluster) string {
	if cluster.Terraform == nil {
		return ""
	}
```

tuunit: i think this would be cleaner but that is just personal preference

src/internal/cmd/generate/generator_test.go
```diff
@@ -0,0 +1,61 @@
...
+		{
+			name: "supported provider is normalized",
+			cluster: config.Cluster{
+				Name: "test-cluster",
+				Terraform: &config.Terraform{
+					Provider: " STACKIT ",
+				},
+			},
+			want: "stackit",
+		},
```
tuunit: this wouldn't load anyways

tuunit: we should really consider refactoring the providers to an enum type and enforcing them more cleanly

tuunit (changes requested):

docs/content/1_getting_started/catalogs.md
Matthiator: Updated the docs: provider-specific directories are Terraform-only. Helm uses conditionals in shared values templates.

src/internal/cmd/generate/generator.go
Matthiator: Removed the optional Helm provider resolution. Terraform validation now uses TerraformProvider.IsSupported().

src/internal/cmd/generate/generator.go
Matthiator: Removed the redundant normalization and now use the typed provider value directly.

src/internal/cmd/generate/generator.go
Matthiator: Removed the resolver function, so the rename no longer applies.

src/internal/cmd/generate/generator_test.go
Matthiator: Added a typed TerraformProvider with centralized validation and removed the impossible test cases.

src/internal/config/types.go
```diff
@@ -4,6 +4,33 @@ import "github.com/kubara-io/kubara/internal/service"
 
 const ConfigVersionV1Alpha1 = "v1alpha1"
 
+// TerraformProvider identifies an infrastructure provider with embedded Terraform templates.
+type TerraformProvider string
+
+const (
+	TerraformProviderPlaceholder TerraformProvider = "<provider>"
+	TerraformProviderStackit     TerraformProvider = "stackit"
```
tuunit: Should we maybe introduce a `none` provider?

`TerraformProviderNone` and use it as the default.

because at the moment if you don't want to use the backend in terraform and don't define the terraform block. You cannot execute `kubara generate` as it will complain that the terraform settings aren't defined. You are forced to use `kubara generate --helm`. This will become messy in the near future

src/internal/config/types.go
```diff
@@ -4,6 +4,33 @@ import "github.com/kubara-io/kubara/internal/service"
 
 const ConfigVersionV1Alpha1 = "v1alpha1"
 
+// TerraformProvider identifies an infrastructure provider with embedded Terraform templates.
+type TerraformProvider string
+
+const (
+	TerraformProviderPlaceholder TerraformProvider = "<provider>"
```
tuunit: Why do we have placeholder enum field?

src/internal/cmd/generate/generator.go
```diff
@@ -54,34 +53,34 @@ func (o *Options) resolveOutputPath(result render.TemplateResult, clusterName st
...
+	if provider == config.TerraformProviderPlaceholder {
 		return "", fmt.Errorf(
 			"cluster %q still uses placeholder provider %q; supported providers: %q",
 			clusterBlock.Name,
 			clusterBlock.Terraform.Provider,
 			supportedProviderList(),
 		)
 	}
-	if !render.SupportedProviders[provider] {
+	if !provider.IsSupported() {
```
tuunit: doesn't `IsSupported` already catch the placeholder case?

src/internal/render/render.go
```diff
@@ -51,6 +45,9 @@ func TemplateFiles(options TemplateOptions) ([]TemplateResult, error) {
 	if err != nil {
 		return nil, fmt.Errorf("get template files for provider %q: %w", options.Provider, err)
 	}
+	if err := rejectHelmProviderPaths(fileList); err != nil {
+		return nil, err
+	}
```
tuunit: do we really need to enforce this? if the user defines the directories they wouldn't be used anyways, right? meaning yes we would render the files but they would not be references by anything. not by argocd, not by any chart or whatsoever.

that is why this feels like unnecessary code to me

tuunit (changes requested):

src/internal/config/types.go
Matthiator: Done. `none` is now the default Terraform provider value and is treated like no Terraform config at runtime. Default `kubara generate` skips Terraform for that cluster, while `kubara generate --terraform` still requires a real provider.

That change touched more than the external-dns template itself: config defaults, schema generation, generator behavior, and tests all need to agree that “no Terraform” is a valid state. I also updated the init update path so existing clusters without Terraform do not panic when they are updated.

src/internal/config/types.go
Matthiator: Removed. The placeholder is no longer part of the provider enum.

src/internal/cmd/generate/generator.go
Matthiator: Yes, agreed. The special placeholder check is gone; unsupported provider values now go through the normal provider validation.

src/internal/render/render.go
Matthiator: Agreed. Removed the guard. Helm `providers/` paths are not treated as provider-specific overrides; if users add such files, they may render, but nothing references them automatically.

src/cmd/generate_test.go
```diff
@@ -402,11 +438,59 @@ func TestGenerateCmd_PlaceholderProviderFailsWithHint(t *testing.T) {
...
+		DockerconfigBase64:          "DockerConfig",
+		ArgocdWizardAccountPassword: "wizardpassword",
+		ArgocdGitHttpsUrl:           "https://example.com",
+		ArgocdGitUsername:           "CoolCapybara",
+		ArgocdGitPatOrPassword:      "password",
+		ArgocdHelmRepoUrl:           "https://example.com",
+		ArgocdHelmRepoUsername:      "CoolCapybara",
+		ArgocdHelmRepoPassword:      "password",
+		DomainName:                  "example.com",
+	})
```
tuunit: we have this at least 4+ times in this test suite please put the test stub generation into a method to reduce the duplication

src/internal/config/types.go
```diff
@@ -4,6 +4,33 @@ import "github.com/kubara-io/kubara/internal/service"
...
+}
+
+// IsSupported reports whether kubara ships Terraform templates for the provider.
+func (p TerraformProvider) IsSupported() bool {
+	for _, supported := range supportedTerraformProviders {
+		if p == supported {
+			return true
+		}
+	}
+	return false
```
tuunit: use new slices.Contains method

tuunit (changes requested):

src/cmd/generate_test.go
Matthiator: Done. Moved the repeated default env fixture into `createDefaultGenerateTestEnv`.

src/internal/config/types.go
Matthiator: Done. `IsSupported` now uses `slices.Contains`.

tuunit (approved):

PR #392, opened 2026-06-11
feat: local init and bootstrap setup for cloud provider independent testing

tuunit:
## 📝 Summary
This change introduces a new "local evaluation" mode for kubara, making it easier for users to try out the platform on their local environment. The main changes include new CLI flags and logic for both `init` and `bootstrap` commands, improved documentation, and a new quick start guide tailored for local evaluation. 

Collectively this hopefully makes it much easier for users to evaluate kubara locally, while keeping production and evaluation workflows clearly separated.

Furthermore, for ease of use this PR makes the following `.env` variables entirely optional and introduces cross validation for the ArgoCD credentials:

* DOCKERCONFIG_BASE64
* ARGOCD_GIT_USERNAME
* ARGOCD_GIT_PAT_OR_PASSWORD

## 🧩 Type of change
- [X] 🔧 CLI / Go code
- [X] 📦 Helm chart
- [ ] 🧱 Terraform module
- [X] 📝 Documentation
- [X] 🧪 Test or CI change
- [X] ♻️ Refactor / cleanup

## 🧪 Testing
- [X] CI passed
- [X] Unit tested
  
## ✅ Checklist


docs/content/1_getting_started/quick_start.md
```diff
@@ -0,0 +1,245 @@
...
+
+It is intended for **local testing only**, not for production use.
+
+This guide supports Linux, MacOS and WSL on Windows.
+
+## Prerequisites
+
+Install kubara first via the [installation guide](installation.md), then make sure these tools are available on your host:
+
+- `sudo` privileges
```
mouhsen-ibrahim: Are you sure that you need `sudo`? it works for me without it

docs/content/1_getting_started/quick_start.md
```diff
@@ -0,0 +1,245 @@
...
+- [Docker](https://docs.docker.com/get-docker/)
```
mouhsen-ibrahim, line 14: Can you please add helm too?

docs/content/1_getting_started/quick_start.md
```diff
@@ -0,0 +1,245 @@
...
+2026-06-11 12:42:02 INF
+🎉 Local kubara bootstrap complete!
+
+📝 Next steps:
+1. Commit and push all repository contents generated by kubara.
+2. Ensure the Git repository is reachable by Argo CD.
+   If you did not provide both ARGOCD_GIT_USERNAME and ARGOCD_GIT_PAT_OR_PASSWORD, ensure the Git repository is public so Argo CD can pull it.
+
+Argo CD should be reachable in a couple of minutes via:
+    https://172.18.0.3.traefik.me/argocd
```
mouhsen-ibrahim, line 121: Where does this IP come from? for me it is 172.20.0.3

docs/content/1_getting_started/quick_start.md
tuunit: @mouhsen-ibrahim on which OS are you?
cloud-provider-kind needs to be started with sudo to be able to listen on privileged 80/443 ports and to attach ips to the loopdevice

docs/content/1_getting_started/quick_start.md
tuunit: Its automatically allocated from the ip range docker reserves on your system. Good point. I'll add an explainer in this guide

docs/content/1_getting_started/quick_start.md
mouhsen-ibrahim: I am using cachyos

src/internal/cmd/bootstrap/local.go
```diff
@@ -0,0 +1,981 @@
...
+func writeOpenBaoKVSecret(ctx context.Context, kubeconfigPath, remoteKey string, fields map[string]string) error {
+	args := []string{
+		"-n", localOpenBaoNamespace,
+		"exec", localOpenBaoPodName,
+		"--",
+		"bao", "kv", "put", fmt.Sprintf("kv/%s", remoteKey),
+	}
+	for key, value := range fields {
+		args = append(args, fmt.Sprintf("%s=%s", key, value))
+	}
```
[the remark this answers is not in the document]
tuunit: this is a local setup for evaluation and not relevant

src/internal/envconfig/store.go
```diff
@@ -118,6 +110,37 @@ func (em *EnvStore) GenerateEnvExample() ([]byte, error) {
...
+		fieldType := t.Field(i)
+
+		if doc := fieldType.Tag.Get("doc"); doc != "" {
+			b.WriteString(doc + "\n")
+		}
+
+		koanfKey := fieldType.Tag.Get("koanf")
+		if koanfKey != "" {
+			fmt.Fprintf(&b, "%s='%v'\n", koanfKey, fieldVal.Interface())
+		}
```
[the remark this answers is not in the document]
tuunit: i think we should enforce naming conventions to follow RFC1123.

but that needs to be done in a separate PR

docs/content/1_getting_started/quick_start.md
tuunit: @mouhsen-ibrahim did you try the full local setup yet?

because yes you can start cloud-provider-kind without sudo but it might fail when later trying to attach an ip

mouhsen-ibrahim: Yes I tried it until the end. I could access ArgoCD and the portal without any issues

tuunit: Perfect thanks for the feedback. I'll look into it on a linux device!

Matthiator (commented):
Thanks for this! The local evaluation flow is really useful, the separation into `localmode` keeps the profile logic nicely contained, and the idempotent bootstrap steps (kind cluster, OpenBao mounts/auth) are much appreciated.

One cross-cutting point I'd like to see called out before merge, plus a few smaller things as inline comments:

**Env validation is relaxed globally, not only for `--local`:** `DOCKERCONFIG_BASE64`, `ARGOCD_GIT_USERNAME`, and `ARGOCD_GIT_PAT_OR_PASSWORD` are now optional for *all* workflows. That may well be intentional (public repos, no registry auth), but it should be mentioned in the PR description and the docs for these variables, because missing registry credentials now silently skip the image pull secret instead of failing validation — a typo in one of these env names only surfaces later as `ImagePullBackOff`. Related: please add cross-field validation for the Git credentials so partial configuration (username without password, or vice versa) fails early with a clear message instead of surfacing as a confusing Argo CD repo connection error.

**Non-blocking follow-ups**, fine to defer (partly depends on #370):
- Align the local OpenBao setup with the model introduced in #370 (KV mount path, auth mount path, namespace-scoped roles, namespace-prefixed secret layout) so both converge on one "OpenBao contract" — might be worth a short ADR.
- Consider standalone OpenBao with file storage instead of dev mode so local state survives Docker/pod restarts (see inline comment on the docs).
- Avoid passing secret values as CLI args to `bao kv put` — stdin via the existing `kubectlExecWithInput` would keep them out of the process lists on the host and in the pod.
- Escape values in the generated `.env` (a single quote in a value currently breaks the file).
- Consider downgrading the hard `sudo` prerequisite check to a warning or doc note.

src/internal/cmd/bootstrap/local.go
```diff
@@ -0,0 +1,983 @@
...
+type localChart struct {
+	DisplayName string
+	Repo        helm.RepoOptions
+	ReleaseName string
+	ChartRef    string
+	Namespace   string
+}
+
+var (
+	localOpenBaoChart = localChart{"OpenBao", helm.RepoOptions{Name: "openbao", URL: "https://openbao.github.io/openbao-helm"}, localOpenBaoReleaseName, "openbao/openbao", localOpenBaoNamespace}
```
Matthiator: The chart is installed as `openbao/openbao` without a pinned version, so an upstream chart release can break `kubara bootstrap --local` without any change on our side. Everything else in kubara is pinned — please pin this too (e.g. add a `Version` field to `localChart` and pass `--version` in `helmUpgradeInstall`).

src/internal/cmd/bootstrap/local.go
```diff
@@ -0,0 +1,983 @@
...
+	}
+
+	if _, err := kubectlExec(ctx, kubeconfigPath,
+		"-n", localOpenBaoNamespace,
+		"exec", localOpenBaoPodName,
+		"--",
+		"bao", "write",
+		fmt.Sprintf("auth/kubernetes/role/%s", localExternalSecretsRole),
+		"bound_service_account_names=*",
+		"bound_service_account_namespaces=*",
```
Matthiator, line 549: Could you add a short comment here that this wide-open role (wildcard service accounts/namespaces with read/list on all of `kv/*`) is a deliberate local-evaluation simplification and not the production model? #370 introduces a least-privilege, namespace-scoped setup for t-cloud-public, and the plan is to converge on one OpenBao contract later — a comment prevents this from being copied as a reference in the meantime.

Tiny related nit: the `configureLocalOpenBao` parameter is named `grafanaAdminPassword` but receives the ArgoCD wizard password — either rename it or note that Grafana intentionally reuses the wizard password locally.

src/internal/cmd/bootstrap/local.go
```diff
@@ -0,0 +1,983 @@
...
+      requests:
+        memory: 512Mi
+      limits:
+        memory: 512Mi
+    volumeMounts:
+      - name: host-ca-certificates
+        mountPath: %s
+        readOnly: true
+    volumes:
+      - name: host-ca-certificates
```
Matthiator, line 729: This hostPath volume (`type: File`) is written unconditionally, while the corresponding kind `extraMounts` entry in `writeLocalKindConfig` is guarded by an `os.Stat` check. On a host without the CA bundle at this path (e.g. macOS, where the stat always fails) the mount is skipped but the repo-server still references the hostPath — it then only works if the kind node image happens to ship its own bundle there. Suggest making this volume conditional on the same check, and/or adding a troubleshooting entry for a repo-server pod stuck on a missing hostPath.

Also note the stat runs on the machine executing kubara, while the mount source is resolved by the Docker daemon (which may live in a VM on macOS/WSL) — so the check can be wrong in both directions; tolerant behavior plus a doc note is probably the most robust combination.

docs/content/1_getting_started/quick_start.md
```diff
@@ -0,0 +1,257 @@
...
+
+1. IMPORTANT: Commit and push all repository contents generated by kubara.
+2. Once more ensure the Git repository is reachable by Argo CD.
+   If you did not provide both ARGOCD_GIT_USERNAME and ARGOCD_GIT_PAT_OR_PASSWORD, ensure the Git repository is public so Argo CD can pull it.
+3. Wait a few minutes until Argo CD syncs the applications from your repository or force a refresh via the UI.
+    - Connect to ArgoCD via port-forwarding and open http://localhost:8080/argocd
+    - Open Argo CD and check whether all apps become `Healthy` and `Synced`. (Maybe force a refresh)
+    - Open Argo CD Settings -> Repositories and check whether Argo can access your repository (is it public? is your pat correct?)
+4. Once the platform components are ready, open the exposed UIs through the local ingress host.
+6. You should be able to reach the kubara Developer Portal via `https://<your-ip>.traefik.me`
```
Matthiator: Few docs nits in this file: the numbered list here jumps from 4 to 6, and several lines carry trailing whitespace (`git diff --check`).

Could you also add a troubleshooting note that OpenBao runs in dev mode and is therefore in-memory: after a Docker or pod restart its config and secrets are gone and the ExternalSecrets start failing — re-running `kubara bootstrap --local <cluster>` restores everything, since the setup is idempotent.

src/internal/catalog/built-in/managed-service-catalog/helm/cert-manager/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.0] - 2026-06-11
+### Changed
```
Matthiator: Nit: this is a new capability, so it belongs under `### Added` rather than `### Changed`.

src/internal/cmd/bootstrap/local.go
tuunit: The kind images are debian based and always ship a /etc/ssl/certs/ca-certificates.crt

tuunit: https://github.com/kubernetes-sigs/kind/blob/main/images/base/Dockerfile

tuunit: Therefore guarding this introduces unnecessary complexity on our side

Matthiator (approved):
lgtm

Issue #346, opened 2026-05-27
[RFC]: Dont generate managed-service-catalog charts for deactivated components


alexander-rabenstein:
### 📝 Summary

As a customer I want to be able to deactivate certain components delivered by kubara. If deactivated kubara shouldn't generate / template the folder under managed-service-catalog/component. 

This way I can keep my repository free from unused helm charts. 

### 🎯 Motivation

Currently If I would for example scan my repository with renovate I would receive PRs even for deactivated components. 
The only way to fix this is to manually cleanup the unneeded components after each kubara run. 

### 💡 Proposal

Add a cli parameter --ignore-deactivated-components to kubara cli to not template the deactivated components. 

### ✅ Checklist

- [x] I have searched existing issues
- [x] This is not a duplicate

tuunit:
This is already planned for one of the next releases :)

tuunit:
Fixed as part of #413 will be released today or tomorrow

PR #413, opened 2026-06-15
feat: add cluster management commands

davrad:
## 📝 Summary
- A new `cluster` command group for the CLI, providing users with subcommands to list and add clusters to their configuration. 
- Some clean up related to the documentation and test utilities to support the new commands. 
- The documentation is updated to reflect the new workflow for adding clusters, and several code improvements are made for consistency and maintainability.

## 🧩 Type of change
- [X] 🔧 CLI / Go code
- [X] 📦 Helm chart
- [ ] 🧱 Terraform module
- [X] 📝 Documentation
- [X] 🧪 Test or CI change
- [X] ♻️ Refactor / cleanup

## 🧪 Testing
- [X] CI passed
- [X] Unit tested
  
## 🔗 Related Issues / Tickets
Addresses #346 

## ✅ Checklist


src/cmd/cluster/add.go
tuunit, line 1: rename to file to simply `add`

src/cmd/cluster/list.go
tuunit, line 1: rename file to simply `list` :)

src/cmd/cluster/cluster.go
```diff
@@ -0,0 +1,16 @@
+package cluster
+
+import "github.com/urfave/cli/v3"
+
+func NewClusterCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "cluster",
+		Usage:       "Manage clusters for Kubara",
```
tuunit:
```suggestion
		Usage:       "Manage your kubara cluster configurations",
```

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,61 @@
...
+func findHubCluster(clusters []config.Cluster) config.Cluster {
+	for _, cluster := range clusters {
+		if cluster.Type == "hub" {
+			return cluster
+		}
+	}
+	// base case, only necessary for compiler,
+	// should not execute in a default environment
+	return config.Cluster{}
+}
```
tuunit:
```suggestion
func findHubCluster(clusters []config.Cluster) *config.Cluster {
	for _, cluster := range clusters {
		if cluster.Type == "hub" {
			return &cluster
		}
	}
	
	return nil
}
```

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,61 @@
...
+
+	configStore := config.NewConfigStoreWithCatalog(configFilePath, catalog.LoadOptions{})
+	err := configStore.Load()
+	if err != nil {
+		return fmt.Errorf("config load: %w", err)
+	}
+	currentConfig := configStore.GetConfig()
+
+	clusters := currentConfig.Clusters
+	hubCluster := findHubCluster(clusters)
```
tuunit:
```suggestion
	hubCluster := findHubCluster(clusters)
	if hubCluster == nil {
		return fmt.Errorf("no hub cluster can be found in config")
	}
```

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,61 @@
...
+
+	//create new cluster and append it
+	newCluster := hubCluster
```
tuunit:
```suggestion
	newCluster := *hubCluster
```

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,61 @@
...
+	err := configStore.Load()
+	if err != nil {
+		return fmt.Errorf("config load: %w", err)
+	}
+	currentConfig := configStore.GetConfig()
+
+	clusters := currentConfig.Clusters
+	hubCluster := findHubCluster(clusters)
+
+	//create new cluster and append it
```
tuunit:
```suggestion
	// create new cluster using hub cluster as its dereferenced template
```

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,61 @@
...
+	newCluster := hubCluster
+	newCluster.Name = spokeName
+	newCluster.Type = "spoke"
```
tuunit: Please introduce an enum for the cluster type in `src/internal/config/types.go` and use that

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,61 @@
...
+
+	spokeServices := maps.Clone(newCluster.Services)
+	newCluster.Services = disableServicesFor(spokeServices, []string{"homer-dashboard", "argocd"})
```
tuunit: mhhh i get why you did this but we need to generalize this using the catalog processing logic and instead of cloning the services from the hub we need to use the catalog provided and apply the defaults from the catalog. especially as some of the services can have a filter that they aren't applicable for spoke cluster types

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"maps"
+
+	"github.com/kubara-io/kubara/internal/catalog"
+	"github.com/kubara-io/kubara/internal/config"
+	"github.com/kubara-io/kubara/internal/service"
+)
+
+func AddCluster(configFilePath string, spokeName string) error {
+
```
tuunit: remove remove empty line

src/internal/cmd/cluster/list.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/kubara-io/kubara/internal/catalog"
+	"github.com/kubara-io/kubara/internal/config"
+)
+
+// Internal Function for the 'kubara cluster ls' command
+// Requires the configFilePath for the ConfigStore
+// Prints out the context in tabular form
+func ListClusters(configFilePath string) error {
+
+	configStore := config.NewConfigStoreWithCatalog(configFilePath, catalog.LoadOptions{})
```
tuunit:
```suggestion
func ListClusters(configFilePath string) error {
	configStore := config.NewConfigStoreWithCatalog(configFilePath, catalog.LoadOptions{})
```

i think this is self explanatory and doesn't need a comment

tuunit (changes requested):

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/urfave/cli/v3"
+)
+
+// Creates the Command for the 'kubara cluster add command
+// Command necessitates a cluster-name as an arg
+func AddCluster() *cli.Command {
+	return &cli.Command{
+		Name:        "add",
+		Usage:       "Create a new spoke cluster for the hub",
+		UsageText:   "kubara cluster add <cluster-name>",
```
tuunit:
```suggestion
		Usage:       "Add a new spoke cluster to your config",
		UsageText:   "kubara cluster add <cluster-name>",
```

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,50 @@
...
+		Description: "Adds a new spoke cluster to the existing hub cluster",
```
tuunit:
```suggestion
		Description: "Adds a new spoke cluster to an existing config yaml",
```

src/cmd/cluster/cluster.go
```diff
@@ -0,0 +1,16 @@
...
+		UsageText:   "kubara cluster [command]",
+		Description: "Enables handling for hub and spoke clusters",
```
tuunit:
```suggestion
		Description: "Enables the configuration and quick setup of clusters",
```

tuunit (changes requested):

src/cmd/cluster/add.go
davrad: Changed filename

src/cmd/cluster/list.go
davrad: Changed filename

src/cmd/cluster/cluster.go
davrad: changed the info

src/internal/cmd/cluster/add.go
davrad: removed code

src/internal/cmd/cluster/add.go
davrad: removed code snippet

src/internal/cmd/cluster/add.go
davrad: removed code snippet

src/internal/cmd/cluster/add.go
davrad: introduced enum

src/internal/cmd/cluster/add.go
davrad: function now applies the catalog to the newly created spoke cluster

src/internal/cmd/cluster/add.go
davrad: removed empty line

src/internal/cmd/cluster/list.go
davrad: I don't think, that it is self explanatory. As someone who has not worked with the library before and who was not familiar with the cmd structure of the project a single comment like this helps explain to me, how the project works with this.
However I do agree that three lines are a bit too much and have removed two lines.

src/cmd/cluster/add.go
davrad: Changed description.

src/cmd/cluster/add.go
davrad: changed description.

src/cmd/cluster/cluster.go
davrad: changed description

src/cmd/cluster/cluster_test.go
```diff
@@ -0,0 +1,75 @@
+package cluster_test
```
tuunit, line 2: Why is this in its own package?

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+
+	"github.com/kubara-io/kubara/internal/catalog"
+	internal "github.com/kubara-io/kubara/internal/cmd/cluster"
+	"github.com/kubara-io/kubara/internal/utils"
+	"github.com/urfave/cli/v3"
+)
+
+// Creates the Command for the 'kubara cluster add command
+// Command necessitates a cluster-name as an arg
```
tuunit:
```suggestion
```

unnecessary comment

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,60 @@
...
+func CreateAddClusterCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "add",
+		Usage:       "Add a new spoke cluster to your config",
+		UsageText:   "kubara cluster add <cluster-name>",
```
tuunit:
```suggestion
		UsageText:   "kubara cluster add CLUSTER_NAME",
```

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,60 @@
...
+			if len(spokeName) == 0 {
+				cli.ShowSubcommandHelpAndExit(cmd, 1)
+			}
+
+			cwd, err := filepath.Abs(cmd.String("work-dir"))
+			if err != nil {
+				return fmt.Errorf("get working directory: %w", err)
+			}
+
+			rawCatalog := strings.TrimSpace(cmd.String("catalog"))
```
tuunit:
```suggestion
			catalogFlag := cmd.String("catalog")
```

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,60 @@
...
+			catalogOptions := catalog.LoadOptions{Overwrite: cmd.Bool("catalog-overwrite")}
+			if rawCatalog != "" {
+				catalogOptions.CatalogPath, err = utils.GetFullPath(rawCatalog, cwd)
+				if err != nil {
+					return fmt.Errorf("get catalog path: %w", err)
+				}
+			}
```
tuunit: note to myself: we need to change this, as https://github.com/kubara-io/kubara/pull/430 will introduce OCI and path support

src/cmd/cluster/cluster_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/kubara-io/kubara/cmd/testutil"
+	"github.com/kubara-io/kubara/internal/catalog"
+	"github.com/kubara-io/kubara/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewClusterCommand(t *testing.T) {
+
+	command := cluster.NewClusterCommand()
```
tuunit, line 16:
```suggestion
func TestNewClusterCommand(t *testing.T) {
	command := cluster.NewClusterCommand()
```

src/cmd/cluster/add.go
tuunit: Instead please add:

```go
			Config: cli.StringConfig{
				TrimSpace: true,
			},
```

to all global StringFlags in `src/cmd/flags.go` e.g.:

```go
		&cli.StringFlag{
			Name:         "file",
			Value:        flags.InputFile,
			Usage:        "Input file path for base64 operation",
			Destination:  &flags.InputFile,
			Config: cli.StringConfig{
				TrimSpace: true,
			},
		},
```

src/internal/catalog/built-in/customer-service-catalog/helm/example/argo-cd/values.yaml.tplt
```diff
@@ -26,6 +26,28 @@ bootstrapValues:
         ignore:
           - kind: Secret
             name: cert-manager-webhook-ca
+{{/*
+Idea: for each spoke cluster we automatically generate/ template an external secret ressource pointing to <cluster-name>/<stage-name>/argocd/<spoke-name>
+TODO: add to docs
```
tuunit: remove and add proper documentation for what this is instead

src/internal/cmd/cluster/add.go
```diff
@@ -0,0 +1,33 @@
+package cluster
+
+import (
+	"fmt"
+
+	"github.com/kubara-io/kubara/internal/catalog"
+	"github.com/kubara-io/kubara/internal/config"
+)
+
+// AddCluster is an internal Function for the 'kubara cluster add [spoke-name]' command
```
tuunit:
```suggestion
```

src/internal/cmd/cluster/add.go
tuunit, line 1: as these two files add and list inside the internal packages are so short. i suggest we move the code to the actual cmd package. not perfect separation of concern but less mental overhead necessary for jumping through the files. i did the same for the catalog commands in #430

src/internal/cmd/generate/generator.go
```diff
@@ -29,8 +30,26 @@ type Options struct {
 	EnvPath            string
 }
 
-// buildTemplateContext creates a map for rendering templates with cluster config, catalog services, and env vars.
-func buildTemplateContext(cluster config.Cluster, cat catalog.Catalog, em envconfig.EnvMap) (map[string]any, error) {
+// getSpokesFromAllClusters returns a list of all spoke Clusters of a given cluster list
+func getSpokesFromAllClusters(clusters []config.Cluster) ([]map[string]any, error) {
```
tuunit:
```suggestion
// getSpokesClusters returns a list of all spoke Clusters of a given cluster list
func getSpokesClusters(clusters []config.Cluster) ([]map[string]any, error) {
```

src/internal/cmd/generate/generator.go
```diff
@@ -29,8 +30,26 @@ type Options struct {
...
+		if cluster.Type != config.Spoke {
+			continue
+		}
+		spokeMap, err := toJSONMap(cluster)
+		if err != nil {
+			return nil, fmt.Errorf("convert spoke %q to map: %w", cluster.Name, err)
+		}
+		spokeMaps = append(spokeMaps, spokeMap)
+
+	}
```
tuunit, line 53:
```suggestion
	for _, cluster := range clusters {
		if cluster.Type != config.Spoke {
			continue
		}
		
		spokeMap, err := toJSONMap(cluster)
		if err != nil {
			return nil, fmt.Errorf("convert spoke %q to map: %w", cluster.Name, err)
		}
		spokeMaps = append(spokeMaps, spokeMap)
	}
```

src/internal/config/factory.go
```diff
@@ -93,3 +93,27 @@ func createServicesFromCatalogWithOptions(catalogOptions catalog.LoadOptions, cl
 
 	return services, nil
 }
+
+// dirty temporyary solution to create blanked spoke cluster
+func CreateBlankSpokeCluster(name string) Cluster {
```
tuunit:
```suggestion
func CreateSpokeScaffolding(name string) Cluster {
```

src/internal/config/factory.go
```diff
@@ -93,3 +93,27 @@ func createServicesFromCatalogWithOptions(catalogOptions catalog.LoadOptions, cl
...
+			Provider:          "<provider>",
+			ProjectID:         "<project-id>",
+			KubernetesType:    "<edge or ske>",
+			KubernetesVersion: "1.34",
+			DNS: DNS{
+				Name:  "<dns-name>",
+				Email: "my-test@nowhere.com",
+			},
+		},
+		Services: nil,
```
tuunit:
```suggestion
		Name:             name,
		Stage:            "<stage>",
		Type:             Spoke,
		DNSName:          "<dns-name>",
		SSOOrg:           "<my-org>",
		SSOTeam:          "<my-team>",
		Terraform: &Terraform{
			Provider:          "<provider>",
			ProjectID:         "<project-id>",
			KubernetesType:    "<edge or ske>",
			KubernetesVersion: "<version>",
			DNS: DNS{
				Name:  "<dns-name>",
				Email: "<dns-email>",
			},
		},
		Services: nil,
```

src/internal/config/types.go
```diff
@@ -70,8 +75,12 @@ type DNS struct {
...
 type ArgoCD struct {
-	Repo     RepoProto       `json:"repo" yaml:"repo" jsonschema:"required,title=ArgoCD Git Repository"`
-	HelmRepo *HelmRepository `json:"helmRepo,omitempty" yaml:"helmRepo,omitempty" jsonschema:"title=ArgoCD Helm Charts Repository"`
+	Repo       RepoProto       `json:"repo" yaml:"repo" jsonschema:"required,title=ArgoCD Git Repository"`
+	HelmRepo   *HelmRepository `json:"helmRepo,omitempty" yaml:"helmRepo,omitempty" jsonschema:"title=ArgoCD Helm Charts Repository"`
+	ClusterRef *ClusterRef     `json:"clusterRef,omitempty" yaml:"clusterRef,omitempty" jsonschema:"title=Cluster Registration,description=Secret backend reference for spoke kubeconfig registration in hub Argo CD."`
+}
+
+type ClusterRef struct {
 }
```
tuunit: Whats the exact goal here?

src/internal/cmd/generate/generator.go
```diff
@@ -180,6 +209,41 @@ func (o *Options) writeTemplateResults(results []render.TemplateResult) error {
...
+			filtered = append(filtered, result)
+			continue
+		}
+		svc := cluster.Services[serviceName]
+		if svc.Status == service.StatusEnabled {
+			filtered = append(filtered, result)
+		}
+	}
+	return filtered
+}
```
tuunit: I think we need to build a more sophisticated solution for this.

We should first of all utilize the field `spec.chartPath` of a service instead of just the name.

Further matching against all pathParts can easily lead to conflicts / false positives. We should really only scan against the third pathPart for managed-service-catalog and the fourth for customer-service-catalog:

- managed-service-catalog/helm/<chart-name>
- customer-service-catalog/helm/test-cluster/<chart-name>

tuunit (changes requested):

src/cmd/cluster/add.go
davrad: Similar reason as to above. If I am a new contributor to the project, as I was I think this comment would have helped me to understand the structure of the cli cmd in the repo is structured.

src/cmd/cluster/add.go
davrad: Updated it

src/cmd/cluster/add.go
davrad: Removed the strings.TrimSpace for the rawCatalog code snippets and added the config to each String flag.

src/internal/cmd/generate/generator.go
davrad: Added and removed empty lines.

src/internal/config/factory.go
davrad: Updated function definition.

src/internal/config/factory.go
davrad: struct now contains dummy values to be changed.

src/internal/config/types.go
davrad: This was a code snippet/ prototype for the argocd spoke cluster templating in the values.yaml. Since it now is templated in a different way, I removed the code above.

src/internal/cmd/generate/generator.go
davrad: Code now returns spec.chartPath for the name and does not iterate over the pathParts themselves anymore, but rather looks at the third or fourth position respectively.

tuunit (dismissed):

src/internal/config/store.go
```diff
@@ -647,6 +644,13 @@ func (cs *ConfigStore) applyServiceCatalogDefaults() error {
 		}
 
 		for name, def := range cat.Services {
+			catalogService := cat.Services[name]
+			supportedTypes := catalogService.Spec.ClusterTypes
+
+			if !slices.Contains(supportedTypes, cluster.Type) {
+				continue
+			}
```
Matthiator: `clusterTypes` is optional, and `factory.go` treats an empty list as "no restriction". Here an omitted `clusterTypes` makes `slices.Contains` return false, so the service is dropped for *every* cluster type. That silently breaks external catalogs that omit the field and is inconsistent with the factory path. Also drops the redundant `catalogService` alias (`def` is already the value).

```suggestion
			if len(def.Spec.ClusterTypes) > 0 && !slices.Contains(def.Spec.ClusterTypes, cluster.Type) {
				continue
			}
```

src/cmd/cluster/list.go
```diff
@@ -0,0 +1,66 @@
...
+			cwd, err := filepath.Abs(cmd.String("work-dir"))
+			if err != nil {
+				return fmt.Errorf("get working directory: %w", err)
+			}
+			configFilePath, err := utils.GetFullPath(cmd.String("config-file"), cwd)
+			if err != nil {
+				return fmt.Errorf("get config file path: %w", err)
+			}
+
+			configStore := config.NewConfigStoreWithCatalog(configFilePath, catalog.LoadOptions{})
```
Matthiator: `cluster list` loads with an empty `LoadOptions{}`, so it ignores `--catalog` and validates external-catalog configs against the built-in schema (can fail before listing). Resolve the catalog flags the same way `add`/`generate` do (`catalog` is already imported here):

```suggestion
			catalogOptions, err := catalog.ResolveLoadOptions(cwd, cmd.String("catalog"), cmd.Bool("catalog-overwrite"))
			if err != nil {
				return fmt.Errorf("could not resolve catalog options: %w", err)
			}

			configStore := config.NewConfigStoreWithCatalog(configFilePath, catalogOptions)
```

src/cmd/testutil/testutil.go
```diff
@@ -0,0 +1,127 @@
...
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/kubara-io/kubara/internal/config"
+	"github.com/kubara-io/kubara/internal/envconfig"
+	"github.com/kubara-io/kubara/internal/service"
+	"github.com/stretchr/testify/require"
+	"github.com/urfave/cli/v3"
+	"go.yaml.in/yaml/v2"
```
Matthiator: This is the only use of `go.yaml.in/yaml/v2` in the tree; everything else (including the original helper this was extracted from) uses `v3`. Switching avoids promoting `v2` to a direct dependency (AGENTS.md: no new deps without need). Run `go mod tidy` afterward to drop the direct `v2` require.

```suggestion
	"go.yaml.in/yaml/v3"
```

Matthiator (commented):
## Review summary

Solid direction — the spoke-cluster workflow, the `ResolveLoadOptions` extraction and the new `testutil` package are clean. Two **High** issues should be resolved before merge; the rest are smaller and posted as inline suggestions.

**High**
1. The new template `PathPredicate` drops the hub's `argo-cd` chart whenever `argocd` is left at its catalog-default `disabled` status (= "not self-managed", not "uninstall"). Verified empirically: with `argocd: disabled`, `generate` emits every chart except `argo-cd`, and `bootstrap` installs Argo CD from exactly that missing path. Regression independent of the new commands. See the inline comment on `generator.go`.
2. `cluster add` writes a config that violates the schema (placeholder enums/formats, empty required `argocd.repo`), so the next `Load()` (in `cluster list`, `generate`, or a second `add`) fails; the test masks it with `_ = configStore.Load()`. Inline suggestions on `factory.go` + `cluster_test.go`.

**Medium**
3. `ApplyServiceCatalogDefaults` now treats an empty `clusterTypes` as "supported nowhere", silently dropping services from external catalogs and diverging from `factory.go` (empty = unrestricted). Inline suggestion on `store.go`.
4. `cluster list` ignores `--catalog`. Inline suggestion on `list.go`.
5. `cluster add` does not reject duplicate or path-unsafe cluster names. Inline suggestion (duplicate check) on `add.go`.

**Low**
6. `testutil` adds `go.yaml.in/yaml/v2` as a new direct dependency while the rest uses `v3`. Inline suggestion on `testutil.go`.
7. `cluster add` prints nothing on success — inline suggestion.
8. `CreateSpokeScaffolding` uses the `"spoke"` literal instead of the new `Spoke` constant — folded into the `factory.go` suggestion.
9. `cluster add` calls `os.Exit` via `ShowSubcommandHelpAndExit` instead of returning — inline suggestion.
10. `cleanupOldFiles` only wipes `managed-service-catalog/{helm,terraform}`, not the per-cluster `customer-service-catalog` overlays. With the new predicate, disabling a service now leaves stale overlay files behind on regeneration; `TestDisabledServicesDontGetWritten` only checks the managed path. Worth either extending cleanup to customer overlays or documenting the limitation and asserting both paths.
11. Removing the `required` services array from the schema is a user-visible change to `kubara schema` output (necessary given the clusterTypes filtering) — worth a note in the docs / an ADR.

src/internal/config/factory.go
```diff
@@ -93,3 +93,25 @@ func createServicesFromCatalogWithOptions(catalogOptions catalog.LoadOptions, cl
...
+			ProjectID:         "<project-id>",
+			KubernetesType:    "<edge or ske>",
+			KubernetesVersion: "<version>",
+			DNS: DNS{
+				Name:  "<dns-name>",
+				Email: "<dns-mail>",
+			},
+		},
+		Services: nil,
+	}
```
Matthiator, line 124: **High (part 1 of 2).** The current placeholders violate the schema: `<provider>` / `<edge or ske>` fail the enums, `<version>` fails the version pattern, `<dns-name>` / `<dns-mail>` fail the hostname/email formats, and the empty `ArgoCD.Repo` fails the required `minProperties=1`. Since `cluster add` does not re-validate after `SaveToFile`, the file is written but the next `Load()` (in `cluster list`, `generate`, or a second `add`) rejects it.

Scaffold schema-valid defaults instead. One concrete option below (tune values to taste; `stackit` is the only currently-supported provider, switch to `TerraformProviderNone` if a spoke should not carry Terraform by default). Also uses the new `Spoke` constant.

```suggestion
	return Cluster{
		Name:    name,
		Stage:   "dev",
		Type:    Spoke,
		DNSName: name + ".example.com",
		SSOOrg:  "<my-org>",
		SSOTeam: "<my-team>",
		Terraform: &Terraform{
			Provider:          TerraformProviderStackit,
			ProjectID:         "<project-id>",
			KubernetesType:    "ske",
			KubernetesVersion: "1.34",
			DNS: DNS{
				Name:  name + ".example.com",
				Email: "platform@example.com",
			},
		},
		ArgoCD: ArgoCD{
			Repo: RepoProto{
				HTTPS: &RepoType{
					Customer: Repository{URL: "https://git.example.com/platform/repo.git", TargetRevision: "main"},
					Managed:  Repository{URL: "https://git.example.com/platform/repo.git", TargetRevision: "main"},
				},
			},
		},
		Services: nil,
	}
```

src/cmd/cluster/cluster_test.go
```diff
@@ -0,0 +1,75 @@
...
+	testutil.CreateDefaultGenerateTestEnv(t, dir)
+
+	cliFlags := flags.NewGlobalFlags().CLIFlags()
+	app := testutil.CreateTestAppWithFlags(cliFlags, cluster.NewClusterCommand())
+	args := []string{"kubara", "--config-file", configPath, "--work-dir", dir, "cluster", "add", spokeName}
+	err := app.Run(context.Background(), args)
+
+	require.NoError(t, err)
+	configStore := config.NewConfigStoreWithCatalog(configPath, catalog.LoadOptions{})
+	_ = configStore.Load()
```
Matthiator: **High (part 2 of 2).** Once the scaffolding produces a valid config, assert the round-trip instead of swallowing the error; otherwise this test passes even when `add` writes an unloadable file (which it does today).

```suggestion
	require.NoError(t, configStore.Load())
```

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,73 @@
...
+
+			configStore := config.NewConfigStoreWithCatalog(configFilePath, catalogOptions)
+			err = configStore.Load()
+			if err != nil {
+				return fmt.Errorf("config load: %w", err)
+			}
+			currentConfig := configStore.GetConfig()
+
+			clusters := currentConfig.Clusters
```
Matthiator, line 58: `add` appends without checking for an existing cluster of the same name, so running it twice creates duplicates. (A name-format check rejecting `/`, `\\`, `.` and `..` is also worth adding to prevent path traversal when the name is later substituted into output paths; that needs a `strings` import so it is left out of this suggestion.)

```suggestion
			clusters := currentConfig.Clusters

			for _, existing := range clusters {
				if existing.Name == spokeName {
					return fmt.Errorf("cluster %q already exists", spokeName)
				}
			}
```

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,73 @@
...
+					TrimSpace: true,
+				},
+			},
+		},
+
+		Action: func(c context.Context, cmd *cli.Command) error {
+			spokeName := cmd.StringArg("cluster-name")
+			if len(spokeName) == 0 {
+				cli.ShowSubcommandHelpAndExit(cmd, 1)
+			}
```
Matthiator, line 34: Returning the help error keeps this consistent with the rest of the command (which returns errors) and avoids an `os.Exit` that the test harness cannot recover from.

```suggestion
			if spokeName == "" {
				return cli.ShowSubcommandHelp(cmd)
			}
```

src/cmd/cluster/add.go
```diff
@@ -0,0 +1,73 @@
...
+			currentConfig.Clusters = append(clusters, newCluster)
+
+			if err = configStore.ApplyServiceCatalogDefaults(); err != nil {
+				return fmt.Errorf("apply spoke catalog defaults: %w", err)
+			}
+			if err = configStore.SaveToFile(); err != nil {
+				return fmt.Errorf("save config to file: %w", err)
+			}
+
+			return nil
```
Matthiator, line 77: `add` succeeds silently. A confirmation line helps, especially since the user still has to edit the scaffolded values afterward.

```suggestion
			fmt.Printf("Added spoke cluster %q to %s\n", spokeName, configFilePath)
			return nil
```

src/internal/cmd/generate/generator.go
```diff
@@ -180,6 +211,44 @@ func (o *Options) writeTemplateResults(results []render.TemplateResult) error {
...
+	chartPathServiceIndex := buildChartPathServiceIndex(cat)
+
+	return func(path string) bool {
+		serviceName := serviceNameFromTemplatePath(chartPathServiceIndex, path)
+		if serviceName == "" {
+			return true
+		}
+
+		svc, ok := cluster.Services[serviceName]
+		return ok && svc.Status == service.StatusEnabled
```
Matthiator, line 253: **High — this drops the hub's Argo CD chart.**

`argocd: disabled` means "Argo CD is not self-managed", not "do not install Argo CD". This predicate treats any non-`enabled` service as "do not render its files", so for a hub with `argocd` at its catalog default (`disabled`) it filters out **both** `managed-service-catalog/helm/argo-cd` **and** the `customer-service-catalog/.../argo-cd` overlay (which also carries this PR's spoke registration). `bootstrap` then installs Argo CD from the now-missing `managed-service-catalog/helm/argo-cd` and fails.

Verified on this branch: with `argocd: disabled`, `generate` emits every service chart **except** `argo-cd`. This regresses the plain `generate` → `bootstrap` flow, independent of the new commands.

The spoke case (a spoke must not get Argo CD) is already handled by the `clusterTypes` filter removing `argocd` from a spoke's service map, so the predicate only needs to stop gating services that must always be rendered.

Suggested direction (keeps the catalog boundary — no hard-coded `"argocd"` in Go): add a catalog metadata flag such as `spec.alwaysRender` (or `bootstrapRequired`) on the service definition, set it on `argo-cd.yaml`, and skip the status gate for those services:

```go
svc, ok := cluster.Services[serviceName]
if cat.Services[serviceName].Spec.AlwaysRender {
	return true
}
return ok && svc.Status == service.StatusEnabled
```

src/internal/config/store.go
tuunit: Indeed 🤔 I wasn't aware that we made it optional. That is rather unfortunate and should maybe be changed. For now I added this defensive check

src/cmd/cluster/list.go
tuunit: fixed

src/cmd/testutil/testutil.go
tuunit: no we don't want to use go.yaml.in at all. it is deprecated we want to use:
 "sigs.k8s.io/yaml"

fixed

src/internal/config/factory.go
tuunit: i do agree with the missing ArgoCD block but the rest will stay as it will fail anyways without intervention/editing of the user. the user needs to decide the values and its just a placeholder. it does not have the goal to work out of the box

src/cmd/cluster/add.go
tuunit: good point

src/cmd/cluster/add.go
tuunit: I don't agree this will stay as we already have the same behaviour in the catalog commands

src/internal/cmd/generate/generator.go
tuunit: As we want to get rid of this weird quirk of ArgoCD anyways. I will hardcode a check for ArgoCD with a TODO to clean this up. ArgoCD should be a fixture of kubara and not of the catalog

tuunit: done

Matthiator (approved):
lgtm

PR #418, opened 2026-06-22
feat!: add namespaced SecretStore support (stores now keyed by map)

Matthiator:
## 📝 Summary

Adds provider-neutral chart support for **namespaced** External Secrets `SecretStore` resources, and aligns the existing cluster-scoped helper to the same shape.

- New helper `templateLibrary.externalSecrets.secretStores` renders one `SecretStore` (scoped to the release namespace) per entry in `.Values.namespacedSecretStores`, wired into the `argo-cd`, `external-dns`, `kube-prometheus-stack`, `oauth2-proxy` and `velero` charts.
- Both `namespacedSecretStores` and `clusterSecretStores` are now keyed by a **map** (the key is the store name) instead of a list, dropping the synthetic `store-<index>` fallback. This addresses review feedback and keeps the two helpers symmetric.
- Dropped the unused `retrySettings` passthrough from both helpers. Nothing sets it, and omitting it just uses ExternalSecrets' built-in retry defaults.

First extracted PR from the split of #370.

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [x] 📦 Helm chart
- [ ] 🧱 Terraform module
- [x] 📝 Documentation
- [ ] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

`clusterSecretStores` (and the new `namespacedSecretStores`) values are now keyed by a map instead of a list. Existing user `additional-values.yaml` files that define `clusterSecretStores` as a list (e.g. the stackit spoke-cluster setup) must migrate to the map shape. See the migration below; the spoke-cluster docs are updated in this PR.

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

- `go build ./...` and full `make test` pass.
- `helm template` smoke test: map values render valid `SecretStore`/`ClusterSecretStore` resources (key used as the name, no `store-<index>`, no `retrySettings`); empty `{}` values render nothing.

## 🔗 Related Issues / Tickets
Part of the split of #370.

## ✅ Checklist

**Migration.** Convert list entries to map keys:

```yaml
# before
clusterSecretStores:
  - name: workload-0-dev
    provider: { ... }

# after
clusterSecretStores:
  workload-0-dev:
    provider: { ... }
```

The same applies to `namespacedSecretStores`. The T Cloud Public values in #370 are moved to the map shape as well.

### DIFF
--- a/src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
+++ b/src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
@@ -1,0 +1,22 @@
+{{- define "templateLibrary.externalSecrets.secretStore" }}
+{{- range $idx, $data := .Values.namespacedSecretStores }}
+{{- $storeName := default (printf "store-%d" $idx) (default $data.storeName $data.name) }}
+apiVersion: external-secrets.io/v1
+kind: SecretStore
+metadata:
+  name: {{ $storeName }}
+  namespace: {{ $.Release.Namespace }}
+  {{- with $data.labels }}
+  labels:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  provider:
+    {{- toYaml $data.provider | nindent 4 }}
+  {{- with $data.retrySettings }}
+  retrySettings:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+---
+{{- end }}
+{{- end }}
--- a/src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
+++ b/src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
@@ -1,10 +1,9 @@
-{{- define "templateLibrary.externalSecrets.secretStore" }}
-{{- range $idx, $data := .Values.namespacedSecretStores }}
-{{- $storeName := default (printf "store-%d" $idx) (default $data.storeName $data.name) }}
+{{- define "templateLibrary.externalSecrets.secretStores" }}
+{{- range $name, $data := .Values.namespacedSecretStores }}
 apiVersion: external-secrets.io/v1
 kind: SecretStore
 metadata:
-  name: {{ $storeName }}
+  name: {{ $name }}
   namespace: {{ $.Release.Namespace }}
   {{- with $data.labels }}
   labels:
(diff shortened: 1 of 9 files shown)


src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  provider:
+    {{- toYaml $data.provider | nindent 4 }}
+  {{- with $data.retrySettings }}
+  retrySettings:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
```
tuunit:
```suggestion
spec:
  {{- toYaml $data.provider | nindent 2 }}
```

how about?

src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
```diff
@@ -0,0 +1,22 @@
+{{- define "templateLibrary.externalSecrets.secretStore" }}
+{{- range $idx, $data := .Values.namespacedSecretStores }}
+{{- $storeName := default (printf "store-%d" $idx) (default $data.storeName $data.name) }}
```
tuunit: how about making this a map instead of a list?

this way the key can automatically be used as the name and we don't need any arbitrary default fallback like "store-<index>"?


```suggestion
{{- define "templateLibrary.externalSecrets.secretStores" }}
{{- range $name, $data := .Values.namespacedSecretStores }}
```

tuunit (changes requested):

src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
Matthiator: Done, switched it to a map keyed by the store name, which drops the `store-<index>` fallback. Renamed the helper to `secretStores` and updated the five `include`s. Did the same to `clusterSecretStores` so both stay consistent.

Heads-up: mapping `clusterSecretStores` is technically breaking for existing list-style `additional-values.yaml` (e.g. the stackit spoke-cluster setup), so I migrated those docs and flagged it in the PR description. The T Cloud Public values in #370 move to the map shape too.

src/internal/catalog/built-in/managed-service-catalog/helm/template-library/templates/external-secrets/_externalSecrets.secretStore.tpl
Matthiator: Kept the explicit `spec.provider` here, it's a schema thing. Flattened it renders:

```yaml
spec:
  vault:        # ExternalSecrets expects this under spec.provider.vault
    ...
```

`spec.vault` isn't a valid `SecretStore`, so I left the structured form in.

Separately, dropped the `retrySettings` block from both helpers: nothing sets it, and leaving it out just uses ExternalSecrets' built-in retry defaults, so it was dead passthrough.

tuunit (approved):

PR #430, opened 2026-06-22
feat: catalog remote distribution

tuunit:
## 📝 Summary

This PR introduces remote catalog distribution, major reworks/refactor the docs so the catalog concept and the overall learning path for platform engineering is easier to follow.

1. Introducing remote catalog distribution using OCI artifacts
  - Added catalog login, catalog pull, and catalog push commands
  - Wired the new remote flow into the existing catalog commands like add, create, and package.
  - Added the backend logic for working with remote catalogs and OCI references.
  - Removed older duplicated command code and kept the behavior in the main command package.
  - Added tests for the new remote/catalog packaging behavior.
2. Reorganizing the docs
  - Split the docs into clearer sections like concepts, building your platform, and workload onboarding.
  - Moved catalog docs out of “getting started” and added new concept/explainer pages for catalogs, catalog distribution, and catalog templating.
  - Updated navigation and overview pages so the new structure is reflected across the docs.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

## 🧪 Testing
- [x] CI passed
- [x] Unit tested
  
## ✅ Checklist


src/internal/catalog/package.go
```diff
@@ -78,11 +79,8 @@ func createCachedArtifact(manifest CatalogManifest, catalogRoot string, artifact
 	}
 
 	packOptions := oras.PackManifestOptions{
-		Layers: []v1.Descriptor{layerDescriptor},
-		ManifestAnnotations: map[string]string{
-			"io.kubara.catalog.name":    manifest.Metadata.Name,
-			"io.kubara.catalog.version": manifest.Spec.Version,
-		},
+		Layers:              []v1.Descriptor{layerDescriptor},
+		ManifestAnnotations: buildCatalogManifestAnnotations(manifest),
```
Matthiator, line 83: **Packaging isn't reproducible -> the cache slowly fills with orphans**

Packaging the *same* untouched catalog twice produces two different digests - I packaged one catalog three times without changing a byte and got three different manifests.

`oras.PackManifest(..., PackManifestVersion1_1, ...)` auto-injects an `org.opencontainers.image.created` timestamp when it isn't present in `ManifestAnnotations`, so the manifest digest drifts on every run even though `TarReproducible = true` keeps the layer itself stable.

Knock-on effects:
- each re-`package`/`pull` re-points the tag ref to a new digest and leaves the old `artifacts/sha256-.../` directory behind. There's no prune/GC (and `ListCachedCatalogs` even builds a `referencedDigests` set it never uses), so the cache grows unbounded over time.
- `pull` then sees byte-identical content as "changed".

Could we set/clear the `created` annotation here so identical input -> identical digest? That makes packaging idempotent and keeps the cache from piling up duplicates. A small `catalog prune` (drop artifact dirs no ref points to) would round it out.

src/internal/catalog/remote.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	cachedArtifact, cachedErr := GetCachedArtifact(ref.Raw)
+	wasCached := cachedErr == nil
+	artifact, err := pullRemoteCatalog(ctx, ref, options.Insecure)
+	if err != nil {
+		return PullResult{}, err
+	}
+	if !ref.IsDigest {
+		if err := writeLocalReference(ref, artifact); err != nil {
```
Matthiator: **`pull` doesn't enforce `tag == spec.version`, but `push` does**

`push` rejects a destination tag that doesn't match the catalog version (the `ref.Tag != artifact.CatalogVersion` check further down in `PushCatalog`), but the pull path writes the local reference keyed by the *requested tag* without ever checking it against the pulled `Catalog.yaml`'s `spec.version`.

So a remote `oci://repo/catalog:1.2.3` whose `Catalog.yaml` says `version: 9.9.9` gets cached under tag `1.2.3` but reports/loads as `9.9.9`. Mirroring the push-side check here would keep the local cache consistent.

src/internal/catalog/remote.go
```diff
@@ -0,0 +1,335 @@
...
+		return PullResult{}, err
+	}
+
+	cachedArtifact, cachedErr := GetCachedArtifact(ref.Raw)
+	wasCached := cachedErr == nil
+	artifact, err := pullRemoteCatalog(ctx, ref, options.Insecure)
+	if err != nil {
+		return PullResult{}, err
+	}
+	if !ref.IsDigest {
```
Matthiator: **Catalogs pulled by digest never show up in `catalog list`**

A local reference is only written for tag pulls (`if !ref.IsDigest`), while `catalog list` walks the `refs/` tree. So after a successful `kubara catalog pull oci://repo/catalog@sha256:...` the artifact is cached and retrievable by digest, but `list` won't show it. Minor, but a little surprising.

src/cmd/catalog/login.go
```diff
@@ -0,0 +1,206 @@
...
+	var err error
+
+	if flags.password && flags.passwordStdin {
+		return internal.LoginOptions{}, fmt.Errorf("password and password-stdin cannot be used together")
+	}
+	if flags.identityToken && flags.identityTokenStdin {
+		return internal.LoginOptions{}, fmt.Errorf("identity-token and identity-token-stdin cannot be used together")
+	}
+
+	usesPasswordAuth := flags.username != "" && (flags.password || flags.passwordStdin)
```
Matthiator: **`login` silently ignores conflicting auth flags**

The guard just below only rejects the combo when password-auth *and* identity-token-auth are both detected - but password-auth is only "detected" when `--username` is set. So e.g. `--identity-token --username alice` or `--identity-token-stdin --password-stdin` fall through to the identity-token path and the conflicting username/password flags get silently ignored instead of rejected. Erroring on any mixed combo would be friendlier.

Matthiator (commented):
Re-reviewed after the latest push (`03f7293b`) and took it for a full end-to-end spin against ghcr.io (`create -> add -> package -> push -> pull -> unpackage`) - works great 🎉 Nice that this push already fixed the first-pull "updated from remote" message and the spelling/grammar nits.

A few things still look worth a look before merge - left the substantive ones as inline comments (the packaging/cache one is the main item, plus a pull/push consistency gap). None are hard security bugs.

A couple of smaller things that don't map cleanly onto a changed line in this diff:

- **Tag policy** (`internal/catalog/types.go`, `StrictCatalogVersion`, enforced in `oci_reference.go`): tags are restricted to strict `MAJOR.MINOR.PATCH`, so prerelease/build tags (`1.2.3-rc.1`, `1.2.3+build`) and `latest` are rejected at parse time. Totally fine if that's the intended policy - might be worth a sentence in the distribution docs, since the example there only says "semver x.y.z".
- **Nits:** `createCachedArtifact` uses `context.Background()` instead of the command context (packaging can't be cancelled); `pull` always hits the network even when the ref is already cached (no offline pull - probably intended as a refresh, just flagging); `sanitizePathComponent` is a denylist (only colon, at-sign and backslash) - safe here since refs are already OCI-validated, but an allowlist would be marginally more robust.

Matthiator (commented):
_Duplicate submission (posted twice by mistake) — please see my other review on this PR for the actual comments._

src/internal/catalog/package.go
tuunit: I agree with the pruning but repacking will produce a new digest which is the same for helm and docker.

tuunit: Otherwise how would we solve this? What would we set the created annotation to when repackaging? If it is already cached we could check against the cache and check whether the blobs did change but what if it isn't cached? Should the package try a pull? And what if the pull fails? Just package a new catalog and overwrite it on push?

this can cause inconsistent behaviour

tuunit: Okay I checked the OCI specification and found out that the annotations are optional therefore I added a couple lines to force immutable manifest digests by setting a fake 1970 date while building/packing the manifest and removing it afterwards before pushing

src/internal/catalog/remote.go
```diff
@@ -0,0 +1,340 @@
...
+
+	cachedArtifact, cachedErr := GetCachedArtifact(ref.Raw)
+	wasCached := cachedErr == nil
+	artifact, err := pullRemoteCatalog(ctx, ref, options.Insecure)
+	if err != nil {
+		return PullResult{}, err
+	}
+	if err := ensureReferenceTagMatchesCatalogVersion(ref, artifact.CatalogVersion, "pull"); err != nil {
+		return PullResult{}, err
+	}
```
Matthiator, line 136: Small edge from the new pull-side check: by this point `pullRemoteCatalog` has already persisted the artifact into the content-addressed cache, so a rejected pull leaves an unreferenced `artifacts/sha256-…` dir behind (until the next `catalog list` prunes it). Since you already have `pruneArtifactIfUnreferenced`, you can clean it up right here:

```suggestion
	if err := ensureReferenceTagMatchesCatalogVersion(ref, artifact.CatalogVersion, "pull"); err != nil {
		// don't leave the just-pulled blob behind as an unreferenced orphan
		_ = pruneArtifactIfUnreferenced(artifact.ManifestDigest)
		return PullResult{}, err
	}
```

(Cleaner alternative: validate the tag inside `pullRemoteCatalog` *before* `finalizeCachedArtifact`, so nothing is written on mismatch — but this is the smaller change.)

src/internal/catalog/package.go
```diff
@@ -78,16 +79,17 @@ func createCachedArtifact(manifest CatalogManifest, catalogRoot string, artifact
...
 	}
 	manifestDescriptor, err := oras.PackManifest(ctx, fileStore, oras.PackManifestVersion1_1, CatalogArtifactType, packOptions)
 	if err != nil {
 		return CachedArtifact{}, fmt.Errorf("pack catalog manifest: %w", err)
 	}
+	// Remove unnecessary org.opencontainers.image.created as we always set it to 1970-01-01 to force immutable
+	// manifest digests for the same contents
+	delete(manifestDescriptor.Annotations, v1.AnnotationCreated)
```
Matthiator: Heads-up: this `delete` doesn't actually affect the published manifest. By the time it runs, `oras.PackManifest` has already serialized the manifest (with `created`), computed the digest from those bytes, and pushed the blob to the store — and `CopyGraph`/push later copy that blob verbatim *by digest*. Deleting from the returned **descriptor's** annotation map doesn't rewrite the blob or change the digest. (Checked it: with only this `delete` and no fixed value, the digest still changes every run and the manifest still carries a real `created`.)

What actually makes the digest reproducible is the fixed `created` value set in `buildCatalogManifestAnnotations` above — which is exactly what the oras docs recommend. So this line can just go:

```suggestion
	}
```

The `1970-01-01` value then simply stays in the manifest, which is the intended reproducible-build behavior.

Matthiator (commented):
Thanks for the quick turnaround 🙏 Re-checked at `a485c50d` — all five points from the previous review are addressed: reproducible digests (fixed `created`), automatic prune of unreferenced artifacts, `pull` now enforces `tag == spec.version`, digest pulls show up in `list`, and the login flag-conflict is rejected. 👍

Two small follow-ups below, neither a blocker.

Matthiator (approved):

PR #432, opened 2026-06-23
fix(renovate): align chart version bump with changelog on major dependency updates

Matthiator:
## 📝 Summary

Fixes a divergence in the Renovate chart-version automation: on a **major** dependency update the CHANGELOG entry and the `Chart.yaml` `version:` field were bumped differently, so they never matched.

`renovate.json` has two mechanisms that bump the umbrella chart's own version when a dependency inside it is updated:

| Mechanism | bumpType for a major dep update |
|---|---|
| `bumpVersions` (writes `Chart.yaml` `version:`) | **minor** (`{{#if isPatch}}patch{{else}}minor{{/if}}`) |
| `postUpgradeTasks` → changelog script | **major** (`{{#if isMajor}}major{{/if}}…`) |

Both read the same pre-bump chart version and bump independently (Renovate runs `postUpgradeTasks` before `bumpVersions`). They agreed on patch/minor but diverged on **major**: `Chart.yaml` got a minor bump (e.g. `1.2.0 → 1.3.0`) while the CHANGELOG got a major bump (`→ [2.0.0]`). Because the umbrella chart's major is always `1`, every major dependency update emitted a phantom `[2.0.0]` header that never matched the real chart version (visible in the `traefik` and `kube-prometheus-stack` CHANGELOGs, and reproduced in #427 / #428).

### Decision

These are **pass-through wrapper charts**: their `values.yaml` re-exposes the upstream chart's value interface directly (the `traefik:` / `kube-prometheus-stack:` block). So an upstream breaking major — e.g. traefik 41 renaming `logs.general` → `log` — propagates to consumers and the wrapper genuinely inherits it. We keep that behaviour (dependency major → chart major) and fix the divergence on the `bumpVersions` side:

```diff
-"bumpType": "{{#if isPatch}}patch{{else}}minor{{/if}}"
+"bumpType": "{{#if isMajor}}major{{else if isMinor}}minor{{else}}patch{{/if}}"
```

Now a dependency major bumps the chart major in **both** places, it sticks in `Chart.yaml`, and the next major goes `2.x → 3.0.0`. The changelog script keeps its original `isMajor` handling unchanged.

**Trade-off:** this fires for any declared Helm dependency in the wrapper `Chart.yaml` (e.g. `prometheus-blackbox-exporter`), not just the chart's namesake — so expect somewhat more frequent major bumps.

## 🧩 Type of change
- [x] 🧪 Test or CI change

## 🧪 Testing
- [ ] CI passed
- [ ] Unit tested
- [x] Not tested (explain why below)

Config-only change to Renovate's bump-type template; the effect is only observable when Renovate next regenerates a major-update PR. Verified by JSON validity and by tracing the divergence against existing CHANGELOG history (the #403 commit confirms both mechanisms bump from the same pre-bump base).

## 🔗 Related Issues / Tickets
Related to #427, #428 — both should be rebased (tick the rebase checkbox) after this merges so their CHANGELOG entry and `Chart.yaml` version line up on the same major.

## ✅ Checklist

The already-merged phantom `[2.0.0]` entries in the `traefik` and `kube-prometheus-stack` CHANGELOGs are left untouched here; they can be cleaned up separately if desired.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -24,12 +24,12 @@
         {
           "filePatterns": ["{{packageFileDir}}/Chart.{yaml,yml}"],
           "matchStrings": ["version:\\s(?<version>[^\\s]+)"],
-          "bumpType": "{{#if isPatch}}patch{{else}}minor{{/if}}"
+          "bumpType": "{{#if isMajor}}major{{else if isMinor}}minor{{else}}patch{{/if}}"
         }
       ],
       "postUpgradeTasks": {
         "commands": [
-          "./.scripts/renovate_bot_update_helm_changelog.sh \"{{{depName}}}\" \"{{{currentVersion}}}\" \"{{{newVersion}}}\" \"{{{packageFileDir}}}\" \"{{#if isPatch}}patch{{else}}minor{{/if}}\""
+          "./.scripts/renovate_bot_update_helm_changelog.sh \"{{{depName}}}\" \"{{{currentVersion}}}\" \"{{{newVersion}}}\" \"{{{packageFileDir}}}\" \"{{#if isMajor}}major{{/if}}{{#if isMinor}}minor{{/if}}{{#if isPatch}}patch{{/if}}\""
         ],
         "fileFilters": ["**/CHANGELOG.md"],
         "executionMode": "update"
--- a/renovate.json
+++ b/renovate.json
@@ -29,7 +29,7 @@
       ],
       "postUpgradeTasks": {
         "commands": [
-          "./.scripts/renovate_bot_update_helm_changelog.sh \"{{{depName}}}\" \"{{{currentVersion}}}\" \"{{{newVersion}}}\" \"{{{packageFileDir}}}\" \"{{#if isMajor}}major{{/if}}{{#if isMinor}}minor{{/if}}{{#if isPatch}}patch{{/if}}\""
+          "./.scripts/renovate_bot_update_helm_changelog.sh \"{{{depName}}}\" \"{{{currentVersion}}}\" \"{{{newVersion}}}\" \"{{{packageFileDir}}}\" \"{{#if isPatch}}patch{{else}}minor{{/if}}\""
         ],
         "fileFilters": ["**/CHANGELOG.md"],
         "executionMode": "update"


renovate.json
```diff
@@ -29,7 +29,7 @@
       ],
       "postUpgradeTasks": {
         "commands": [
-          "./.scripts/renovate_bot_update_helm_changelog.sh \"{{{depName}}}\" \"{{{currentVersion}}}\" \"{{{newVersion}}}\" \"{{{packageFileDir}}}\" \"{{#if isMajor}}major{{/if}}{{#if isMinor}}minor{{/if}}{{#if isPatch}}patch{{/if}}\""
+          "./.scripts/renovate_bot_update_helm_changelog.sh \"{{{depName}}}\" \"{{{currentVersion}}}\" \"{{{newVersion}}}\" \"{{{packageFileDir}}}\" \"{{#if isPatch}}patch{{else}}minor{{/if}}\""
```
tuunit: but do we want to get rid of the isMajor?

Matthiator: Good question. We actually built it this way deliberately in the past. But I've now thought about it for a while longer, and I think we shouldn't drop `isMajor`.

These are pass-through wrappers: `values.yaml` re-exposes the upstream chart's value interface directly (the `traefik:` / `kube-prometheus-stack:` block). So when upstream ships a breaking major — e.g. traefik 41 renaming `logs.general` → `log` — that breaking change propagates to anyone configuring it through our wrapper. By SemVer the wrapper inherits the major, which is exactly what `isMajor` encodes.

The real bug is the divergence: `bumpVersions` (line 27) maps `major → minor`, while this script receives `major → major`. Both read the same pre-bump chart version and bump independently (Renovate runs `postUpgradeTasks` before `bumpVersions`), so on a major dep update `Chart.yaml` goes e.g. `1.2.0 → 1.3.0` but the CHANGELOG entry becomes `[2.0.0]`. Since the chart major is always `1`, every major dep update emits a phantom `[2.0.0]` that never matches the real version.

So instead of dropping `isMajor`, I'll align the two by keeping this line and changing line 27 to:

```
"{{#if isMajor}}major{{else if isMinor}}minor{{else}}patch{{/if}}"
```

Then a dep major → chart major in both places, it sticks in `Chart.yaml`, and the next major goes `2.x → 3.0.0`.

One trade-off worth flagging: this fires for any declared Helm dependency in the wrapper `Chart.yaml` (e.g. `prometheus-blackbox-exporter` in kube-prometheus-stack), not just the chart's namesake — so expect somewhat more frequent major bumps. That's consistent with "a major may carry breaking changes," but noting it.

tuunit (approved):

PR #434, opened 2026-06-25
feat: add kubara agents command for coding-agent onboarding

Matthiator:
## 📝 Summary

More and more people drive infrastructure work through coding agents (Claude Code, Codex, …). Today, an agent spinning up a platform with kubara has to crawl the full docs site to orient itself — that burns a lot of tokens, is slow, and is error-prone. This PR is a first step toward **first-class coding-agent support** in kubara: helping agents work with the framework faster, with far fewer tokens, and giving us a credible "optimized for coding agents" story that matches where the ecosystem is heading.

It adds a new opt-in command:

```
kubara agents [--overwrite]
```

It writes a single `AGENTS.md` into the working directory — the file Claude Code / Codex auto-read — so the agent gets a compact, token-lean entry point instead of reading dozens of rendered HTML pages.

**Design choices (kept deliberately low-maintenance):**
- **Delegates instead of duplicating.** `AGENTS.md` points at the self-describing CLI (`kubara --help`, `kubara schema`) for command and config details, so that content can't drift. Only the stable workflow, repo conventions, and a small curated doc map are hand-written.
- **Version-pinned raw Markdown.** Documentation links target the raw `.md` sources on GitHub pinned to the installed binary version (raw Markdown is much cheaper for an agent to read than rendered HTML; release builds pin to the tag, dev/snapshot builds fall back to `main`).
- **Opt-in, not wired into `init`.** Nothing is forced on users who don't use agents. Skip-if-exists, refreshed with `--overwrite`.
- **Drift guard with correct locality.** A test asserts every referenced `docs/content/...md` path exists. It skips when the docs tree is absent (src-only sparse CI checkout) and runs in the `docs-check` workflow, so a documentation move that breaks a link fails *its own* PR with a clear message, instead of surprising an unrelated Go PR later.

`llms.txt` was intentionally kept out of the repo for now — its natural home is the hosted docs site (`docs.kubara.io/llms.txt`), which can be generated at docs-build time in a follow-up.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

No. Purely additive: a new command plus its docs; no existing behavior changes.

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

Ran `kubara agents` locally and inspected the rendered `AGENTS.md` (version pinning + skip/overwrite behavior). Unit tests cover version-ref resolution, write/skip/overwrite, and the doc-link drift guard. Also verified `go build`, `go vet`, `gofmt`, and `golangci-lint` (0 issues). The command does not touch a cluster.

## ✅ Checklist

Possible follow-ups (out of scope here): generate `docs.kubara.io/llms.txt` at docs-build time for the web/standard discovery path, and a `kubara validate`/doctor command that flags unfilled `change-me` values in generated artifacts — the next big token/backtracking saver for both agents and humans.


src/internal/agentcontext/templates/AGENTS.md.tmpl
```diff
@@ -0,0 +1,60 @@
...
+
+Use `kubara generate --dry-run` to preview without writing.
+
+## What lives in this repository
+
+- `config.yaml` — the platform definition (clusters, stages, services). Run `kubara schema` for its
+  JSON Schema instead of guessing field names.
+- `.env` — inputs/credentials consumed by `kubara init`. Git-ignored; never commit it.
+- `customer-service-catalog/` — generated Helm values and Terraform. Search for `change-me`
+  placeholders: those are the values to fill in before bootstrapping.
```
tuunit: I don't think we have change-me placeholders anymore

src/internal/agentcontext/templates/AGENTS.md.tmpl
```diff
@@ -0,0 +1,60 @@
...
+- Authoritative config reference: run `kubara schema` (emits the JSON Schema for `config.yaml`)
+- Rendered docs (latest): {{ .DocsSite }}
+
+## Core workflow (run in this order)
+
+1. `kubara init` — generate `config.yaml` from `.env`.
+2. `kubara generate --helm` (add `--terraform` if you provision infrastructure with kubara) — render
+   Helm/Terraform artifacts from the service catalog into this repository.
+3. `kubara bootstrap <cluster-name>` — install Argo CD onto the target cluster.
+4. Commit & push the generated files. Argo CD reconciles everything else from Git.
```
tuunit: Maybe add a first step like:

1. Ensure you are in a git repository that is either empty or already contains a config.yaml

src/internal/agentcontext/templates/AGENTS.md.tmpl
```diff
@@ -0,0 +1,60 @@
...
+- `customer-service-catalog/` — generated Helm values and Terraform. Search for `change-me`
+  placeholders: those are the values to fill in before bootstrapping.
+
+## Conventions for agents
+
+- The CLI is fully non-interactive. Drive it with flags, the `.env` file, or `KUBARA_*` environment
+  variables — never expect an interactive prompt.
+- Reach for `kubara <command> --help` and `kubara schema` before inferring flags or config fields.
+- Don't hand-write Argo CD `Application`s or chart values from scratch; let the catalog and
+  `kubara generate` produce them, then edit the generated `change-me` values.
```
tuunit: This is a wrong we should explicitly encourage proper usage and consulting the user to evaluate what is the best solution given the circumstances:

* Use Application/AppSets for workloads that are cluster specific
* Use the catalog when component is part of the reusable platform

https://docs.kubara.io/next/2_concepts/catalogs/#when-not-to-use-a-catalog

tuunit (changes requested):

src/internal/agentcontext/templates/AGENTS.md.tmpl
Matthiator: Confirmed — verified there are no live `change-me` placeholders in the generated overlays anymore (only changelog/commented references remain), so I reworded this to "review and fill in any environment-specific values". Also fixed the stale `change-me` section in `bootstrapping.md`. Done in 5d04db8e.

src/internal/agentcontext/templates/AGENTS.md.tmpl
Matthiator: Added as the new first workflow step: "Make sure you are in a git repository that is either empty or already contains a `config.yaml`." (remaining steps renumbered). Done in 5d04db8e.

src/internal/agentcontext/templates/AGENTS.md.tmpl
Matthiator: Good point — rewrote this to match the tool to the workload and consult the user when unclear: Argo CD `Application`/`ApplicationSet`s for cluster-specific workloads, and the catalog for reusable platform components (pointing at the "when not to use a catalog" guidance). Done in 5d04db8e.

tuunit (approved):

PR #461, opened 2026-06-29
ci: image collection and scanning

davrad:
## 📝 Summary
This PR adds image scanning and Trivy analysis to the CI.
In addition it updates the goreleaser schema to include an image section into the header, such that in each release, all the image versions of the standard kubara catalog are also listed.

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [x] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [x] CI passed
- [ ] Unit tested
  
## ✅ Checklist

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -426,3 +426,31 @@ jobs:
           name: terraform-validation-${{ github.run_attempt }}-${{ github.sha }}-${{ github.run_number }}
           retention-days: ${{ env.ARTIFACT_RETENTION_DAYS }}
           path: ${{ env.REPORT_DIR }}/terraform
+
+
+  extract-images:
+    needs: run-kubara-init-and-generate
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Checkout CI scripts
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          sparse-checkout: |
+            .scripts
+            .github/helm-profiles
+
+      - name: Download generated kubara artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: kubara-generated-files-${{ github.run_attempt }}-${{ github.sha }}-${{ github.run_number }}
+
+      - name: Verify Helm
+        run: helm version --short
+
+      - name: Extract container images
+        run: |
+          mkdir -p reports
+          OUTPUT_FILE=reports/images.txt .scripts/image-version.sh
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -431,6 +431,8 @@ jobs:
   extract-images:
     needs: run-kubara-init-and-generate
     runs-on: ubuntu-latest
+    outputs:
+      images: ${{ steps.matrix.outputs.images }}
     defaults:
       run:
         shell: bash
@@ -471,3 +473,44 @@ jobs:
           name: container-images-${{ github.run_attempt }}-${{ github.sha }}-${{ github.run_number }}
           retention-days: ${{ env.ARTIFACT_RETENTION_DAYS }}
           path: reports/images.txt
+    
+      - name: Build image matrix
+        id: matrix
+        run: |
+          set -euo pipefail
+          images=$(jq -R -s -c 'split("\n") | map(select(length > 0))' reports/images.txt)
+          echo "images=$images" >> "$GITHUB_OUTPUT"
+          if [[ "$images" == "[]" ]]; then
+            echo "has_images=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_images=true" >> "$GITHUB_OUTPUT"
+          fi
+    
+  trivy-scan:
+    needs: extract-images
+    if: needs.extract-images.outputs.has_images == 'true'
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 10
+      matrix:
+        image: ${{ fromJSON(needs.extract0images.outputs.images) }}
+      steps:
+        - name: Scan ${{ matrix.image }}
+          uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+          with:
+            image-ref: ${{ matrix.image }}
+            version: ${{ env.TRIVY_VERSION }}
+            scanners: vuln
+            severity: CRITICAL # high as well?
+            ignore-unfixed: true
+            format: SARIF
+            output: trivy-results.sarif
+            exit-code: '0'
+
+        - name: Upload SARIF
+          if: always()
+          uses: github/codeql-action/upload-sarif@c35d1b164463ee62a100735382aaaa525c5d3496 # v2.25.6
+          with:
+            sarif_file: trivy-results.sarif
+            category: trivy-${{ matrix.image }}
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -454,3 +454,20 @@ jobs:
         run: |
           mkdir -p reports
           OUTPUT_FILE=reports/images.txt .scripts/image-version.sh
+
+      - name: Write image summary
+        if: always()
+        run: |
+          [[ -f reports/images.txt ]] || exit 0
+          echo "### Container Images" >> "$GITHUB_STEP_SUMMARY"
+          while read -r img; do
+            echo "- $img" >> "$GITHUB_STEP_SUMMARY"
+          done < reports/images.txt
+
+      - name: Upload image list
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        if: always()
+        with:
+          name: container-images-${{ github.run_attempt }}-${{ github.sha }}-${{ github.run_number }}
+          retention-days: ${{ env.ARTIFACT_RETENTION_DAYS }}
+          path: reports/images.txt
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -433,6 +433,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       images: ${{ steps.matrix.outputs.images }}
+      has_images: ${{ steps.matrix.outputs.has_images }}
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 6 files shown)


.scripts/image-version.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# pipefail that pipes break
+set -euo pipefail
+
+export PATH="$HOME/.local/bin/:$PATH"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+MANAGED="${MANAGED:-${PWD}/managed-service-catalog/helm}"
+PROFILES="${PROFILES:-${SCRIPT_DIR}/../.github/helm-profiles}"
+OUTPUT_FILE="${OUTPUT_FILE:-}"
```
tuunit, line 14: I think the current profiles don't activate all components. We need to render all our charts. At the moment the report is missing metrics-server, longhorn, loki, external-dns, metallb, velero

davrad:
~The Pipepiline currently runs into Caching issues.~

Caching issues were addressed.

.scripts/image-version.sh
davrad: All images are now rendered, and checked by Trivy :)

.scripts/kubara-config-update.sh
```diff
@@ -49,4 +49,14 @@ apply_yaml_if_set KUBARA_KUBERNETES_VERSION  ".clusters[0].terraform.kubernetesV
...
 
+# Enable every catalog service to the image/vuln report and the release header cover all charts
+yq eval '(.clusters[0].services[] | .status) = "enabled"' -i "$CFG"
+
+# metalb, loki and velero need custom configs
+yq eval '.clusters[0].services.metallb.config.publicLoadBalancerIPs = "203.0.113.10"' -i "$CFG"
+yq eval '.clusters[0].services.metallb.config.loadBalancerAddressPool = ["203.0.113.0/24"]' -i "$CFG"
+yq eval '.clusters[0].storage.bucketNames.chunks = "loki"' -i "$CFG"
+
+yq eval '.clusters[0].services.velero.config.backupStorage.s3Url = "https://bucket.example.com"' -i "$CFG"
```
tuunit, line 60: this will effect all pipelines, which me probably don't want 🤔 as i don't know what the side effects will be

tuunit (changes requested):

.scripts/kubara-config-update.sh
tuunit: okay is only used in workflows/pr-checks.yaml

.scripts/kubara-config-update.sh
```diff
@@ -49,4 +49,14 @@ apply_yaml_if_set KUBARA_KUBERNETES_VERSION  ".clusters[0].terraform.kubernetesV
 apply_yaml_if_set KUBARA_DNS_NAME            ".clusters[0].dnsName"
 apply_yaml_if_set KUBARA_DNS_NAME            ".clusters[0].terraform.dns.name"
 
+# Enable every catalog service to the image/vuln report and the release header cover all charts
+yq eval '(.clusters[0].services[] | .status) = "enabled"' -i "$CFG"
```
Matthiator, line 53: I think the intent is to enable everything so all charts get rendered for the scan — makes sense 👍

The catch: this same generated config is also consumed by the existing `helm-checks` job, which now renders `reloader` (disabled by default) and fails on its `template-library` dependency → the whole job goes red (that's the currently failing check).

Could we scope the "enable all" to a separate config used only for the extract/scan step, so `helm-checks` keeps its current service set? (Alternatively: make `reloader` renderable, or only enable services that render cleanly.)

Matthiator (commented):
Really nice addition — rendering every chart and scanning all catalog images is a great step 🙌

I went through it in detail and left inline comments. The main thing before merge is that `helm-checks` currently goes red (see the config comment), plus a few path/error-handling edge cases in the release job and the extraction script. Nothing here is hard to fix. Happy to help with any of them!

A couple of tiny nits I didn't inline: typo `helm template for for '$chart'` (double "for"), the comment `# pipefail that pipes break` reads a bit confusingly, the config comment `Enable every catalog service to the image/vuln report…` is missing a word, and `SCRIPT_DIR` looks unused. All optional. 🙂

.github/workflows/release.yaml
```diff
@@ -57,6 +57,52 @@ jobs:
...
+          mkdir -p /tmp/gen
+          (cd src && go run main.go --work-dir /tmp/gen init --prep)
+          .scripts/kubara-env-update.sh /tmp/gen/.env
+          (cd src && go run main.go --work-dir /tmp/gen init)
+          .scripts/kubara-config-update.sh /tmp/gen/config.yaml
+          (cd src && go run main.go --work-dir /tmp/gen generate)
+
+      - name: Extract images and set goreleaser header
+        run: |
+          MANAGED=/tmp/gen/managed-service-catalog/helm \
```
Matthiator, line 96: Small path mismatch here: the generated files live under `/tmp/gen`, but only `MANAGED` is overridden — and it points to `managed-service-catalog/helm`, while `generate` now produces `platform-components/helm`. `CONFIG_FILE`/`CONFIGS` also stay relative to the repo root, where there's no `config.yaml`.

As written the script would exit before GoReleaser, so the header wouldn't be filled. Something like:
```
MANAGED=/tmp/gen/platform-components/helm \
CONFIG_FILE=/tmp/gen/config.yaml \
CONFIGS=/tmp/gen/platform-configs/<cluster>/helm \
  "$GITHUB_WORKSPACE"/.scripts/image-version.sh
```
should fix it. (Not caught by PR CI since release only runs on tag push.)

.scripts/image-version.sh
```diff
@@ -0,0 +1,84 @@
...
+if [[ "$PROMETHEUS_STATUS" == enabled ]]; then
+  HELM_TEMPLATE_ARGS+=(--api-versions "monitoring.coreos.com/v1")
+fi
+
+
+echo "Rendering charts from $MANAGED (kube-version=$KUBE_VERSION)" >&2
+
+
+render_dir="$(mktemp -d)"; trap 'rm -rf "$render_dir"' EXIT
+FAILED=()
```
Matthiator, line 38: Nice that failures get collected in `FAILED` — but the array is never checked afterwards, so a chart that fails `dependency update`/`template` is silently skipped and the job still passes. That means its images quietly disappear from the scan (this is exactly what happens to `reloader`).

Could we fail (or at least emit a visible `::warning::`) when `FAILED` is non-empty?

Minor: on the `helm dependency update` line below, `dep_out=$(… >/dev/null 2>&1)` discards all output, so the later `echo "$dep_out"` never prints anything.

.github/workflows/pr-checks.yaml
```diff
@@ -426,3 +428,136 @@ jobs:
...
+
+
+      - name: Scan ${{ matrix.image }}
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        with:
+          cache: false # uses restored db
+          image-ref: ${{ matrix.image }}
+          version: ${{ env.TRIVY_VERSION }}
+          scanners: vuln
+          severity: HIGH
```
Matthiator, line 551: `severity: HIGH` only reports HIGH — `CRITICAL` findings are excluded from the SARIF. Did you mean `severity: HIGH,CRITICAL`? (Combined with `exit-code: '0'` the scan is report-only, which is totally fine if that's intended.)

.scripts/image-version.sh
```diff
@@ -0,0 +1,84 @@
...
+    grep -E '^[[:space:]]*image:' |
+    sed -E "s/^[[:space:]]*image:[[:space:]]*//; s/[\"']//g" |
+    grep -vE '[*!]' |          # drop kyverno wildcard/negation entries
+    grep -vE '^[[:space:]]*$' |
+    sort -u
+)"
+
+echo "Done Rendering!"
+
+[[ -n "$IMAGES" ]] || { echo "::warning::No image references found"; exit 0; }
```
Matthiator, line 77: With `set -euo pipefail`, if `grep` matches nothing the pipeline exits 1 and the script dies in the `IMAGES="$( … )"` assignment above — before ever reaching this friendly `::warning::No image references found` / `exit 0`. `OUTPUT_FILE` then never gets written, and the workflow reads `reports/images.txt` unconditionally, so the `has_images=false` branch can't actually be reached.

A `|| true` on the grep (or writing the file first) would restore the intended empty-case handling.

.scripts/image-version.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# pipefail that pipes break
+set -euo pipefail
+
+export PATH="$HOME/.local/bin/:$PATH"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+MANAGED="${MANAGED:-${PWD}/platform-components/helm}"
+CONFIG_FILE="${CONFIG_FILE:-config.yaml}"
+CLUSTER_NAME="$(yq -r '.clusters[0].name' "$CONFIG_FILE")"
```
Matthiator, line 12: Tiny thing: `yq` reads `$CONFIG_FILE` here before the `[[ -f … ]]` / `command -v` checks below, so a missing file/tool produces a cryptic command-substitution error instead of your nice diagnostic. Moving the checks above this line would surface the friendly message.

.scripts/kubara-config-update.sh
```diff
@@ -49,4 +49,14 @@ apply_yaml_if_set KUBARA_KUBERNETES_VERSION  ".clusters[0].terraform.kubernetesV
...
+
+# metalb, loki and velero need custom configs
+yq eval '.clusters[0].services.metallb.config.publicLoadBalancerIPs = "203.0.113.10"' -i "$CFG"
+yq eval '.clusters[0].services.metallb.config.loadBalancerAddressPool = ["203.0.113.0/24"]' -i "$CFG"
+yq eval '.clusters[0].storage.bucketNames.chunks = "loki"' -i "$CFG"
```
Matthiator, line 58: This writes `.clusters[0].storage.bucketNames.chunks`, but I couldn't find a matching field in the config schema (the Loki chunks bucket comes from the static `platform-components/helm/loki/values.yaml`). It may be silently ignored — worth double-checking whether this actually has the intended effect.

.scripts/image-version.sh
```diff
@@ -0,0 +1,84 @@
...
+            > "$render_dir/$chart.yaml" 2> "$render_dir/$chart.err"; then
+        echo "::error::helm template for for '$chart':"
+        sed 's/^/    /' "$render_dir/$chart.err" >&2
+        FAILED+=("$chart:template"); continue
+    fi
+done
+
+IMAGES="$(
+  cat "$render_dir"/*.yaml |
+    grep -E '^[[:space:]]*image:' |
```
Matthiator, line 68: Quick note on coverage: since extraction is a `grep image:` over the `helm template` output, it catches all standard container / init-container / sidecar images (good!), but by design it misses:
- runtime-injected images (mutating webhooks, Kyverno, operator-set images via flags),
- image refs passed via `args`/`env` rather than an `image:` field,
- `- image:` list-item lines — the regex only matches `image:` after whitespace, not after `- `.

Not a blocker, but might be worth a comment noting this is best-effort, so "scan passed" isn't read as "every running image was scanned."

tuunit:
closed in favour of #501

PR #462, opened 2026-06-30
docs: agent friendly documentation extension

tuunit:
## 📝 Summary
This PR adds automatic generation of a table of contents in markdown format and adds the original markdown files to the generated docs for AI agent processing.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [X] 📝 Documentation
- [ ] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## ⚠️ Is this a breaking change?
- [ ] Yes, this change breaks existing functionality (explain in summary)

## 🧪 Testing
- [ ] CI passed
- [X] Manually tested (local/dev cluster)
- [ ] Unit tested
- [ ] Not tested (explain why below)
  
## ✅ Checklist
- [X] Code compiles and passes all tests
- [ ] Linting and style checks pass
- [ ] Comments added for complex logic
- [X] Documentation updated (if applicable)

PR #463, opened 2026-06-30
fix: restore in-body llms.txt discovery hint for coding agents

Matthiator:
## 📝 Summary

Follow-up to #462. Coding agents (e.g. Copilot) fetch a docs page and read the **body** text, dropping `<head>` — so the `<link rel="alternate">` head pointer from #462 isn't surfaced, and they crawl individual pages instead of fetching `llms.txt`.

This restores the in-body discovery hint (a visually-hidden block linking the index) so agents find it, with two fixes over the original #462 version:
- **relative href** instead of root `/llms.txt` — correct under mike's `/<version>/` deploys (root `/llms.txt` 404s there);
- **a11y-clean** — the hint is machine-only, so it's `aria-hidden` and the link is `tabindex="-1"` (a clipped but focusable link would otherwise be an invisible keyboard tab stop). The misleading `sr-only` class is dropped.

The `<head>` `<link rel="alternate">` stays as a bonus for structured crawlers.

## 🧩 Type of change
- [x] 📝 Documentation

- [ ] Yes, this change breaks existing functionality

No.

## 🧪 Testing

`mkdocs build --strict` passes; verified the hint renders on every page with a relative href (`../../llms.txt` by depth, `llms.txt` on home), `aria-hidden`, and `tabindex="-1"`. Docs-only change, no cluster involved.

## 🔗 Related Issues / Tickets
Follow-up to #462.

## ✅ Checklist

### DIFF
--- a/docs/content/stylesheets/extra.css
+++ b/docs/content/stylesheets/extra.css
@@ -58,5 +58,6 @@ blockquote[data-agent-docs-index="true"] {
   overflow: hidden;
   white-space: nowrap;
   clip: rect(0, 0, 0, 0);
+  clip-path: inset(50%);
   border: 0;
 }
--- a/docs/content/stylesheets/extra.css
+++ b/docs/content/stylesheets/extra.css
@@ -47,3 +47,16 @@
 .mermaid {
     text-align: center;
 }
+/* Visually hide the agent documentation-index hint while keeping it in the
+   rendered text (so coding agents that read the body still find it). */
+blockquote[data-agent-docs-index="true"] {
+  position: absolute;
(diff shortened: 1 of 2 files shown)


Matthiator:
Pushed an a11y hardening from review: the in-body hint is machine-only content, so it's now `aria-hidden` and the link is `tabindex="-1"` — a clipped but focusable link is otherwise still an invisible keyboard tab stop. Also dropped the dangling `sr-only` class (hiding is driven by the data attribute) and added `clip-path`. `mkdocs build --strict` clean.

tuunit (approved):

PR #465, opened 2026-07-01
feat!: migrate catalogs and GitOps repos to the new platform layout

tuunit:
## 📝 Summary

1. Renames and restructures the catalog layout to the new platform naming:
   - `managed-service-catalog` -> `platform-components`
   - `customer-service-catalog` -> `platform-configs`
   - removes extra nesting such as `providers/` and `example/`
   - renames `values.yaml.tplt` to `values.generated.yaml.tplt`
2. Adds migration support for existing setups:
   - migrates `v1alpha2` config and GitOps repo structure to `v1alpha3`
   - migrates `additional-values.yaml` to `values-additional.yaml`
   - migrates existing Helm values files into the new layout
3. Moves migration logic into an isolated package and updates existing tests for the new structure.

GitOps repo structure changes:
```
managed-service-catalog/ -> platform-components/
customer-service-catalog/{helm,terraform,...}/{cluster-1,cluster-xxx} -> platform-configs/{cluster-1,cluster-xxx}/{helm,terraform,...}
additional-values.yaml -> values-additional.yaml
```

Catalog structure changes:
```
managed-service-catalog/helm/ -> platform-components/helm/
managed-service-catalog/terraform/providers/{stackit,xxx} -> platform-components/terraform/{stackit,xxx}
customer-service-catalog/{helm,terraform,...}/example -> platform-configs/{helm,terraform,...}
values.yaml.tplt -> values.generated.yaml.tplt
```

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

- [x] Yes, this changes the catalog and GitOps directory/file naming, but includes migrations for existing setups.

## 🧪 Testing
- [x] CI passed
- [x] Unit tested
  
## ✅ Checklist


la-cc:
`helm-checks.sh` still validates against the old `values.yaml`. It should use `values.generated.yaml` instead.

Please update the following lines in `helm-checks.sh`:

```bash
values_file="$CONFIGS/$chart/values.yaml"
```

to:

```bash
values_file="$CONFIGS/$chart/values.generated.yaml"
```

Affected lines: 106 and 211.

la-cc:
The issue seems to be related to Argo CD’s manifest cache and the current `manifest-generate-paths` annotation:

```yaml
argocd.argoproj.io/manifest-generate-paths: .;..
```

This only watches the chart path, e.g. `platform-components/helm/traefik`, but not the values under `platform-configs/.../helm/traefik`.

As a result, changes to `values.generated.yaml` or additional values files may not invalidate the manifest cache during a normal refresh. Argo CD only picks them up after a hard refresh.

I updated the annotation to also include the platform config values path:

```yaml
argocd.argoproj.io/manifest-generate-paths: ".;..;/{{ $localCtx.platformConfigs.path }}/{{ `{{name}}` }}/helm/{{ $app.path }}"
```

Tested with changes to `values.generated.yaml` and additional values files. Argo CD now detects them without requiring a hard refresh.


_argo.appset.tpl

src/internal/config/migrations/v1alpha2.go
```diff
@@ -0,0 +1,284 @@
...
+		closeErr := root.Close()
+		if err == nil && closeErr != nil {
+			err = fmt.Errorf("close root %q: %w", cwd, closeErr)
+		}
+	}()
+
+	var renameAdditional []string
+	var renameLegacyValues []string
+
+	if err := fs.WalkDir(root.FS(), ".", func(path string, entry fs.DirEntry, walkErr error) error {
```
Matthiator, line 272: The walk starts at `cwd` (which is `--work-dir`, not necessarily the GitOps repo root) and decides what to rename based only on the parent directory name. Two side effects come out of that:

1. It can touch files outside the catalog. Any `values.yaml` / `additional-values.yaml` whose parent folder happens to be named like a built-in service gets renamed — including files in unrelated projects below the working directory. This also runs on read-only-feeling commands like `cluster list`, since the migration happens in `ConfigStore.Load()`.

2. It also renames the components-side chart defaults (`managed-service-catalog/helm/<service>/values.yaml`). That file is meant to stay `values.yaml` — it's static, ships in the catalog, and the ApplicationSet references it explicitly as `- "values.yaml"`. After the rename that reference silently resolves to nothing (`ignoreMissingValueFiles`), so the chart deploys without its bundled defaults. Per the PR description only the configs-side template (`values.yaml.tplt` → `values.generated.yaml.tplt`) and `additional-values.yaml` should change.

Both renames only ever apply to the configs tree, so scoping the walk to the configs catalog root (`customer-service-catalog` / `platform-configs`) instead of `cwd` would fix both at once — it keeps the components `values.yaml` untouched and can't reach anything outside the catalog.

Minor: the rename loop below still returns `fmt.Errorf("remove %q: …")` — leftover wording from before 63b2461c.

src/internal/config/migrations/v1alpha2.go
```diff
@@ -0,0 +1,284 @@
...
+func migrateV1Alpha2Files(cwd string, clusterName string) error {
+	if clusterName == "" {
+		return fmt.Errorf("clusterName is empty")
+	}
+
+	if err := migrateLegacyValuesFiles(cwd); err != nil {
+		return err
+	}
+
+	pattern := filepath.Join(cwd, "*", "*", clusterName)
```
Matthiator: Same root cause as the walk above: `*/*/<clusterName>` isn't anchored to the catalog dirs, so it matches anything two levels deep under `cwd`. Two ways that bites:

- if a cluster is named like a built-in service (e.g. `argo-cd`), the glob also matches `managed-service-catalog/helm/argo-cd` — a chart, not a cluster dir — and moves it, breaking the chart.
- anything unrelated under `--work-dir` shaped like `foo/bar/<clusterName>` gets moved too.

Since the cluster restructure only ever applies to the configs side (`managed-service-catalog` is just renamed flat, no cluster level), scoping this to `customer-service-catalog` / `platform-configs` would avoid both.

src/internal/cmd/generate/generator.go
```diff
@@ -226,9 +225,9 @@ func serviceNameFromTemplatePath(chartPathServiceIndex map[string]string, path s
 	}
 
 	switch {
-	case pathParts[0] == render.DefaultManagedCatalogPath && pathParts[1] == render.Helm.String():
+	case pathParts[0] == render.DefaultPlatformComponentsPath && pathParts[1] == render.Helm.String():
 		return chartPathServiceIndex[pathParts[2]]
-	case len(pathParts) >= 5 && pathParts[0] == render.DefaultOverlayValuesPath && pathParts[1] == render.Helm.String():
+	case len(pathParts) >= 5 && pathParts[0] == render.DefaultPlatformConfigsPath && pathParts[1] == render.Helm.String():
```
Matthiator: This configs case still assumes the old `example/` layout. After dropping `example/`, a configs values path is `platform-configs/helm/<service>/values.generated.yaml.tplt` — 4 segments, service at index 2. So `len(pathParts) >= 5` with `pathParts[3]` never matches, and `serviceNameFromTemplatePath` returns `""` for every configs template. The enabled-service predicate then treats those as non-service files and renders them even for disabled services (the argocd special-case is bypassed too). Should be `>= 4` / `pathParts[2]`, matching the components case above.

src/internal/config/types.go
```diff
@@ -42,7 +43,7 @@ func SupportedTerraformProviders() []TerraformProvider {
 
 // Config is the root of the configuration structure.
 type Config struct {
-	Version  string    `json:"version,omitempty" yaml:"version,omitempty" jsonschema:"title=Config Version,description=The schema version of this config file.,enum=v1alpha2,default=v1alpha2"`
+	Version  string    `json:"version,omitempty" yaml:"version,omitempty" jsonschema:"title=Config Version,description=The schema version of this config file.,enum=v1alpha3,default=v1alpha3"`
```
Matthiator, line 46: The schema now defaults to `v1alpha3` here, but `SaveToFile` (store.go) still defaults an empty version to `v1alpha1`. Since `init` never sets `Version`, a freshly created config is written to disk as `v1alpha1` and then runs the full migration chain — including the v1alpha2→v1alpha3 filesystem migration — on the next command that loads it. The `SaveToFile` default should match the current schema version.

Matthiator (commented):
Went through the whole PR — the Go code in detail (especially the new `migrations` package), and skimmed the template/directory renames. Overall this looks solid: the migration chain is cleanly extracted into its own package, and the naming + restructure is applied consistently across code, tests and docs.

I left 4 inline comments I'd suggest looking at before merge:

- **`migrateLegacyValuesFiles`** — the rename walk starts at `cwd` and matches on the parent-dir name alone, so it can touch files outside the catalog (and renames the components-side chart `values.yaml`, which is meant to stay).
- **`migrateV1Alpha2Files`** — the `*/*/<clusterName>` glob isn't anchored to the catalog dirs, so it can move unrelated dirs / collide with a chart named like the cluster.
- **`SaveToFile` default version** (commented on `types.go`) — fresh configs are written as `v1alpha1`, so a new setup runs the full migration chain on the next command.
- **`serviceNameFromTemplatePath`** — the configs case still assumes the old `example/` layout, so disabled services still get their `values.generated.yaml` rendered.

The first two share the same root cause (scope the migration to the known catalog roots). The rest of the PR looks good to me.

src/internal/config/migrations/v1alpha2.go
tuunit: 1. highly unlikely especially with the current user base. i don't think it is worth the additional effort for such an edge case for a one off migration
2. / 3. i decided to scope the migration to customer-service-catalog -> platform-configs and managed-service-catalog -> platform-components and add a statement in the upcoming changelog of the next release that people who have used different direction names using the generation flags. First need to rename those folders to customer-service-catalog and managed-service-catalog before running kubara next time

tuunit: Last but not least I decided to get rid of the support for custom directory names through generation time only flags.

https://github.com/kubara-io/kubara/pull/465/changes/554807cc480dc499c66b1e673d73ce4de572fb57

src/internal/config/migrations/v1alpha2.go
tuunit: fixed by explicitly searching for filepath.Join(cwd, "customer-service-catalog", "*", clusterName)

src/internal/cmd/generate/generator.go
tuunit: resolved

src/internal/config/types.go
tuunit: fixed

Matthiator (approved):
lgtm

PR #481, opened 2026-07-07
feat: remove dns configuration from .env 

fabian-schmitt:
## 📝 Summary

  

We wanted to change the default behaviour for dns-names which generated a name from .env values by appending project-name.stage.domain.

Now it will be manually configured in config.yaml both for "clusters[].dnsName" and "clusters[].terraform.dns.name"
Pipeline will also set the value manually and not assume there is some logic generating these fields.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [ ] Unit tested
  
## ✅ Checklist

### DIFF
--- a/src/internal/config/factory.go
+++ b/src/internal/config/factory.go
@@ -9,7 +9,7 @@ import (
 )
 
 func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.LoadOptions) (Cluster, error) {
-	dnsName := e.ProjectName + "-" + e.ProjectStage + "." + e.DomainName
+	dnsName := e.ProjectName + "." + e.DomainName
 	services, err := createServicesFromCatalogWithOptions(catalogOptions, "")
 	if err != nil {
 		return Cluster{}, fmt.Errorf("create services from catalog: %w", err)
--- a/src/internal/config/factory.go
+++ b/src/internal/config/factory.go
@@ -9,7 +9,6 @@ import (
 )
 
 func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.LoadOptions) (Cluster, error) {
-	dnsName := e.ProjectName + "." + e.DomainName
 	services, err := createServicesFromCatalogWithOptions(catalogOptions, "")
 	if err != nil {
 		return Cluster{}, fmt.Errorf("create services from catalog: %w", err)
@@ -40,7 +39,7 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
 		Name:             e.ProjectName,
 		Stage:            e.ProjectStage,
 		Type:             "<hub or spoke>",
-		DNSName:          dnsName,
+		DNSName:          "<for example project.my-domain.com>",
 		SSOOrg:           "<my-org>",
 		SSOTeam:          "<my-team>",
 		IngressClassName: "traefik",
--- a/src/internal/config/factory.go
+++ b/src/internal/config/factory.go
@@ -49,7 +49,7 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
 			KubernetesType:    "<edge, ske or cce>",
 			KubernetesVersion: "1.34",
 			DNS: DNS{
-				Name:  dnsName,
+				Name:  "<same as DNSName above>",
 				Email: "my-test@nowhere.com",
 			},
 		},
--- a/src/internal/config/factory.go
+++ b/src/internal/config/factory.go
@@ -39,7 +39,7 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
 		Name:             e.ProjectName,
 		Stage:            e.ProjectStage,
 		Type:             "<hub or spoke>",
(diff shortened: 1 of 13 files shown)


src/internal/envconfig/env.go
```diff
@@ -34,9 +34,7 @@ type EnvMap struct {
 	_                           struct{} `doc:"\n# Project related values"`
 	ProjectName                 string   `default:"<...>" koanf:"PROJECT_NAME"`
 	ProjectStage                string   `default:"<...>" koanf:"PROJECT_STAGE"`
-	_                           struct{} `doc:"\n# DNS Name/Zones related values"`
-	_                           struct{} `doc:"# The Domain name under which your dns-entries will be added."`
-	_                           struct{} `doc:"# The resulting dnsZone name will be a concatenation of <PROJECT_NAME>-<PROJECT_STAGE>.<DOMAIN_NAME>"`
+	_                           struct{} `doc:"\n# DNS Name/Zones related value"`
 	_                           struct{} `doc:"# the value should be looking like 'stackit.zone' eg. 'yourDomain.com'"`
 	DomainName                  string   `default:"<...>" koanf:"DOMAIN_NAME"`
```
tuunit: so we don't use this anymore at all?

src/internal/config/factory.go
```diff
@@ -40,7 +39,7 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
 		Name:             e.ProjectName,
 		Stage:            e.ProjectStage,
 		Type:             "<hub or spoke>",
-		DNSName:          dnsName,
+		DNSName:          "<subdomain.my-domain.com>",
```
tuunit, line 42: Shouldn't this use the domain name from the config instead?


```suggestion
		DNSName:          e.DomainName,
```

src/internal/config/factory.go
```diff
@@ -50,7 +49,7 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
 			KubernetesType:    "<edge, ske or cce>",
 			KubernetesVersion: "1.34",
 			DNS: DNS{
-				Name:  dnsName,
+				Name:  "<subdomain.my-domain.com>",
```
tuunit, line 52:
```suggestion
				Name:  e.DomainName,
```

src/internal/workflow/orchestrator_test.go
```diff
@@ -55,8 +55,8 @@ func TestCreateOrUpdateClusterFromEnv_UpdatesExistingClusterIncludingHelmRepo(t
 	require.Len(t, cfg.Clusters, 1)
 	updated := cfg.Clusters[0]
 	assert.Equal(t, "dev", updated.Stage)
-	assert.Equal(t, "kubara-test-dev.example.com", updated.DNSName)
-	assert.Equal(t, "kubara-test-dev.example.com", updated.Terraform.DNS.Name)
+	assert.Equal(t, "kubara-test-stage.example.com", updated.DNSName)
+	assert.Equal(t, "kubara-test-stage.example.com", updated.Terraform.DNS.Name)
```
tuunit, line 58: why this change?

src/internal/workflow/orchestrator.go
```diff
@@ -18,10 +18,6 @@ func CreateOrUpdateClusterFromEnvWithCatalog(cfg *config.Config, e *envconfig.En
 
 			// Apply the new values from the environment to the found cluster.
 			cfg.Clusters[i].Stage = e.ProjectStage
-			cfg.Clusters[i].DNSName = dnsName
-			if cfg.Clusters[i].Terraform != nil {
-				cfg.Clusters[i].Terraform.DNS.Name = dnsName
-			}
```
tuunit, line 24: Why even add "KUBARA_DNS_NAME" to the test case above and the remove the assignment here?

Shouldn't it be?

```go
			cfg.Clusters[i].DNSName = e.DomainName
			if cfg.Clusters[i].Terraform != nil {
				cfg.Clusters[i].Terraform.DNS.Name = e.DomainName
			}
```

tuunit (changes requested):

src/internal/workflow/orchestrator_test.go
fabian-schmitt: The tests expects the name of the cluster to change - it won't since there is no logic to append stage to subdomain etc.

src/internal/config/factory.go
fabian-schmitt: In this case an init --overwrite with set env variable (legacy behaviour / legacy .env) would overwrite a now to be manually constructed / set DNSName. So then .env values > config.yaml but we want it the other way round... i tried to create a "minimaly invasive change"

src/internal/workflow/orchestrator.go
fabian-schmitt: my approach would be to set this fully manual and have absolutely now interference with legacy .env / overwrite behvaiour like mentioned above.

src/internal/envconfig/env.go
fabian-schmitt: yes

tuunit (approved):

PR #483, opened 2026-07-09
feat: add argocd managed namespaces support and user token auth for cluster secrets

tuunit:
## 📝 Summary
Adds a small improvement to the Helm template for our cluster external secrets. 

* Adding a conditional block that includes the `namespaces` field in the generated cluster secret manifest when `.bootstrapValues.cluster[].namespaces` is set.
* Adding a conditional block to allow for using the user token in addition to the certificate based authentication

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [X] 📦 Helm chart
- [ ] 🧱 Terraform module
- [X] 📝 Documentation
- [ ] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [X] CI passed
- [ ] Unit tested
  
## ✅ Checklist

### DIFF
--- a/src/internal/catalog/built-in/platform-components/helm/cert-manager/Chart.yaml
+++ b/src/internal/catalog/built-in/platform-components/helm/cert-manager/Chart.yaml
@@ -7,7 +7,7 @@ version: 0.4.0
 dependencies:
   - name: template-library
     repository: file://../template-library
-    version: 0.1.0
+    version: 0.2.0
   - name: cert-manager
     version: v1.21.0
     repository: https://charts.jetstack.io
--- a/src/internal/catalog/built-in/platform-components/helm/cert-manager/Chart.yaml
+++ b/src/internal/catalog/built-in/platform-components/helm/cert-manager/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 name: cert-manager
 description: Umbrella Chart for cert-manager
 type: application
-version: 0.4.0
+version: 0.5.0
 dependencies:
   - name: template-library
     repository: file://../template-library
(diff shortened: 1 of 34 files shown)


Matthiator:
A few small review thoughts from my side:

**Token/auth intention**

Small clarification question: the PR description says the token is used in addition to certificate-based auth, and the template keeps `certData`/`keyData` while adding `bearerToken`.

Is that the intended behavior? If yes, maybe `authMethod: token` is a bit misleading because it sounds like token-only auth. A short note in the docs, or a more explicit option name, would make this easier to understand.

**Namespaces docs**

Tiny docs suggestion: `namespaces` is rendered with `join ","`, so it needs to be a YAML list.

Could we add a small example like this?

```yaml
bootstrapValues:
  cluster:
    - name: spoke-dev
      namespaces:
        - team-a
        - team-b
```

**authMethod typos**

Small robustness thought: every `authMethod` value except exactly `token` currently falls back to the certificate branch.

Might be worth documenting the accepted values, or failing on unknown values, so typos like `Token` or `tokne` do not silently change the behavior.

**CHANGELOG / version**

Since this changes the `template-library` behavior, could we also add a CHANGELOG entry and bump the chart version if that is the convention here?

tuunit:
@Matthiator addressed your concerns

src/internal/catalog/built-in/platform-components/helm/cert-manager/Chart.yaml
```diff
@@ -7,7 +7,7 @@ version: 0.4.0
 dependencies:
   - name: template-library
     repository: file://../template-library
-    version: 0.1.0
+    version: 0.2.0
```
Matthiator, line 10: Small versioning note: this dependency bump changes the chart content, but the chart version and CHANGELOG stay unchanged. Could we bump/document the consuming charts as well, like we did for `argo-cd`?

Matthiator (commented):
Small versioning note inline.

src/internal/catalog/built-in/platform-components/helm/cert-manager/Chart.yaml
tuunit: done

Matthiator (approved):
lgtm

PR #490, opened 2026-07-10
fix(external-dns): publish the public IP via default-targets

Matthiator:
## 📝 Summary

When a cluster's load balancer service holds a private address (e.g. a MetalLB pool IP) while the cluster is reached through a routed public IP, external-dns published the **private** address as the DNS target. The Traefik dashboard `IngressRoute` had no per-resource target override, so the cluster hostname ended up with an unreachable private A record next to the public one — the record set round-robins and the dashboard becomes intermittently unreachable.

This sets external-dns `default-targets` to the configured `publicLoadBalancerIPs` together with `force-default-targets`, so every published record points at the public IP regardless of source (`ingress` and `traefik-proxy`). It replaces the per-ingress `external-dns.alpha.kubernetes.io/target` annotations — and the `metallb` status branching they required — which are removed from the service value templates (argo-cd, homer-dashboard, kube-prometheus-stack, kyverno-policy-reporter, longhorn, oauth2-proxy).

The block is guarded by `publicLoadBalancerIPs`: clusters without it (e.g. SKE, where the LB IP is already public) render unchanged.

## 🧩 Type of change
- [ ] 🔧 CLI / Go code
- [x] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [ ] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

Rendered `kubara generate` for both paths: with `metallb` enabled + `publicLoadBalancerIPs` set, external-dns gets `default-targets` + `force-default-targets` and the ingress values carry no target annotation; without it, the block is omitted and output is unchanged. `go test ./internal/render/... ./internal/cmd/generate/...` passes.

## ✅ Checklist

Observed on an Edge Cloud cluster: `dig` returned both the public IP and the private MetalLB pool IP for the ingress host; the private record came from the `traefik-proxy` source publishing the Traefik service's (private) load balancer IP for the dashboard `IngressRoute`.

### DIFF
--- a/src/internal/catalog/built-in/platform-configs/helm/external-dns/values.generated.yaml.tplt
+++ b/src/internal/catalog/built-in/platform-configs/helm/external-dns/values.generated.yaml.tplt
@@ -66,6 +66,16 @@ external-dns:
   extraArgs:
     source:
       - traefik-proxy
+    {{- $publicLoadBalancerIP := dig "cluster" "services" "metallb" "config" "publicLoadBalancerIPs" "" . }}
+    {{- if $publicLoadBalancerIP }}
+    # The load balancer service holds a private address (e.g. a MetalLB pool
+    # IP) while the cluster is reached through a routed public IP. Publish that
+    # public IP for every record, overriding the private target the sources
+    # (ingress and traefik-proxy) would otherwise derive from the service.
+    default-targets:
+      - {{ $publicLoadBalancerIP }}
+    force-default-targets: true
+    {{- end }}
     webhook-provider-url: http://localhost:8080
   rbac:
     create: true
--- a/src/internal/catalog/built-in/platform-configs/helm/external-dns/values.generated.yaml.tplt
+++ b/src/internal/catalog/built-in/platform-configs/helm/external-dns/values.generated.yaml.tplt
@@ -66,14 +66,19 @@ external-dns:
   extraArgs:
     source:
       - traefik-proxy
-    {{- $publicLoadBalancerIP := dig "cluster" "services" "metallb" "config" "publicLoadBalancerIPs" "" . }}
-    {{- if $publicLoadBalancerIP }}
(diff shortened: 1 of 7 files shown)


src/internal/catalog/built-in/platform-configs/helm/external-dns/values.generated.yaml.tplt
```diff
@@ -66,6 +66,16 @@ external-dns:
...
+    {{- $publicLoadBalancerIP := dig "cluster" "services" "metallb" "config" "publicLoadBalancerIPs" "" . }}
+    {{- if $publicLoadBalancerIP }}
+    # The load balancer service holds a private address (e.g. a MetalLB pool
+    # IP) while the cluster is reached through a routed public IP. Publish that
+    # public IP for every record, overriding the private target the sources
+    # (ingress and traefik-proxy) would otherwise derive from the service.
+    default-targets:
+      - {{ $publicLoadBalancerIP }}
+    force-default-targets: true
+    {{- end }}
```
tuunit: publicLoadBalancerIPs can contain multiple ips, separated by comma, will this be supported with this syntax?

Matthiator: Good catch — it wasn't. `publicLoadBalancerIPs` is a comma-separated string, and `- {{ $publicLoadBalancerIP }}` emitted it as a single list entry (one invalid comma-joined target). Fixed in 94f57ba: it now splits on comma and emits one trimmed `default-targets` entry per IP (empty segments skipped). Verified via `kubara generate` — `"192.0.2.50, 198.51.100.10"` renders as two separate targets.

tuunit (approved):

PR #501, opened 2026-07-17
ci: link pinned catalog releases

davrad:
## 📝 Summary

This PR adds the default bootstrap and general catalog releases to the Kubara release notes.

The release workflow reads the catalog versions pinned by Kubara, verifies that the matching catalog releases exist, and passes their versioned links to GoReleaser. This keeps the detailed image inventories with the independently versioned catalog releases and avoids depending on mutable GHCR artifacts.

The existing 1.1.0 catalog releases use the legacy tag format without `v`; newer versions use the current `catalog-vX.Y.Z` convention.

## 🧩 Type of change

- [ ] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

## 🧪 Testing

- [ ] CI passed
- [x] Manually tested
- [ ] Unit tested

Validated locally with `actionlint`, pre-commit, `goreleaser check`, and requests against both currently pinned catalog releases.

## 🔗 Related Issues / Tickets

- Related catalog release work: kubara-io/catalogs#20

## ✅ Checklist

- [x] Code compiles and passes relevant checks

Projects can override the default catalog versions in their configuration; the release notes intentionally show the defaults shipped with this Kubara version.


Matthiator (commented):
Nice — a transparent image inventory in the release notes + as a CI artifact is a great idea 🙌 (I understand Trivy scanning is being handled separately, so I've kept that out of scope here and only flagged the leftover scaffolding as cleanup.)

A few things before merge, mainly two: the collection currently doesn't actually write its output file (so the job is green but produces nothing), and the release header step has a few path issues. Details inline.

Minor nits I didn't inline: typo `helm template for for '$chart'` (double "for"), the `# pipefail that pipes break` comment reads a bit confusingly, `SCRIPT_DIR` looks unused, and the velero `"12.1.0"` → `12.1.0` change is a no-op (still a YAML string). The `reloader` → `template-library` 0.2.0 bump looks correct 👍

.github/workflows/pr-checks.yaml
```diff
@@ -426,3 +428,52 @@ jobs:
...
+        with:
+          name: kubara-generated-files-${{ github.run_attempt }}-${{ github.sha }}-${{ github.run_number }}
+
+      - name: Verify Helm
+        run: helm version --short
+
+      - name: Extract container images
+        run: |
+          mkdir -p reports
+          OUTPUT_FILE="$PWD/reports/images.txt" .scripts/image-version.sh
```
Matthiator: This passes `OUTPUT_FILE`, but the script now reads `IMAGE_OUTPUT_FILE` — so `reports/images.txt` is never written. The current run confirms it: the summary stays empty and the upload logs `##[warning]No files were found with the provided path: reports/images.txt`. The job is green but produces nothing. Renaming the env var to `IMAGE_OUTPUT_FILE` (and `HELM_IMAGE_OUTPUT_FILE` if you also want the dependency list) fixes it.

.github/workflows/release.yaml
```diff
@@ -57,6 +57,52 @@ jobs:
...
+          mkdir -p /tmp/gen
+          (cd src && go run main.go --work-dir /tmp/gen init --prep)
+          .scripts/kubara-env-update.sh /tmp/gen/.env
+          (cd src && go run main.go --work-dir /tmp/gen init)
+          .scripts/kubara-config-update.sh /tmp/gen/config.yaml
+          (cd src && go run main.go --work-dir /tmp/gen generate)
+
+      - name: Extract images and set goreleaser header
+        run: |
+          MANAGED=/tmp/gen/managed-service-catalog/helm \
```
Matthiator: Three path issues in this step:
- `MANAGED` points to `managed-service-catalog/helm`, but `generate` now produces `platform-components/helm`.
- `CONFIG_FILE`/`CONFIGS` are unset → they default relative to the repo root, but the generated `config.yaml` lives under `/tmp/gen`.
- same `OUTPUT_FILE` vs `IMAGE_OUTPUT_FILE` mismatch as in pr-checks → `/tmp/images.txt` is never created.

As written the script exits before GoReleaser, so the header wouldn't be filled. Something like:
```
MANAGED=/tmp/gen/platform-components/helm \
CONFIG_FILE=/tmp/gen/config.yaml \
CONFIGS=/tmp/gen/platform-configs/<cluster>/helm \
IMAGE_OUTPUT_FILE=/tmp/images.txt \
  "$GITHUB_WORKSPACE"/.scripts/image-version.sh
```
(Not caught by PR CI since release only runs on tag push.)

.scripts/image-version.sh
```diff
@@ -0,0 +1,95 @@
...
+if [[ "$PROMETHEUS_STATUS" == enabled ]]; then
+  HELM_TEMPLATE_ARGS+=(--api-versions "monitoring.coreos.com/v1")
+fi
+
+
+echo "Rendering charts from $MANAGED (kube-version=$KUBE_VERSION)" >&2
+
+
+render_dir="$(mktemp -d)"; trap 'rm -rf "$render_dir"' EXIT
+FAILED=()
```
Matthiator: Failures get collected in `FAILED`, but the array is never checked afterwards — a chart that fails `dependency update`/`template` is silently skipped and the job still passes, so its images quietly disappear from the list. Could we fail (or emit a visible `::warning::`) when `FAILED` is non-empty?

Minor: on the `helm dependency update` line below, `dep_out=$(… >/dev/null 2>&1)` discards all output, so the later `echo "$dep_out"` never prints anything.

.scripts/image-version.sh
```diff
@@ -0,0 +1,95 @@
...
+    sort -u
+)"
+
+echo "Done Rendering!"
+
+[[ -n "$IMAGES" ]] || { echo "::warning::No image references found"; exit 0; }
+
+echo "$IMAGES"
+echo "Extracting Helm Dependencies"
+HELM_IMAGES="$(find . -name Chart.yaml -exec yq '.dependencies[] | select(.name != "template-library") | .name + ": " + .version' {} \;)"
```
Matthiator: A few things on the dependency collection:
- `find .` isn't scoped to `$MANAGED` — in the release job (full repo checkout) this scans the entire built-in catalog, not just the rendered set, so it's inconsistent between pr-checks and release. `find "$MANAGED" …` would scope it.
- these are chart `name: version` pairs (Helm **chart** versions, not container images), so the `HELM_IMAGES` naming is a bit misleading.
- `HELM_IMAGE_OUTPUT_FILE` is never set by either workflow, so this list is computed but never persisted anywhere.

.scripts/image-version.sh
```diff
@@ -0,0 +1,95 @@
...
+echo "$HELM_IMAGES"
+
+[[ -n "$HELM_IMAGES" ]] || { echo "::warning::No helm image references found"; exit 0; }
```
Matthiator: This `exit 0` on empty helm-deps runs *before* the `IMAGE_OUTPUT_FILE` write below — so a perfectly valid container-image list wouldn't be written if no dependencies are found. Worth reordering so the image file is written first.

.github/workflows/pr-checks.yaml
```diff
@@ -426,3 +428,52 @@ jobs:
           name: terraform-validation-${{ github.run_attempt }}-${{ github.sha }}-${{ github.run_number }}
           retention-days: ${{ env.ARTIFACT_RETENTION_DAYS }}
           path: ${{ env.REPORT_DIR }}/terraform
+
+
+  extract-images:
+    needs: run-kubara-init-and-generate
+    runs-on: ubuntu-latest
+    outputs:
+      images: ${{ steps.matrix.outputs.images }}
```
Matthiator: Leftover from the earlier Trivy version: these `outputs.images` / `has_images` reference `steps.matrix`, which no longer exists in this job, and `TRIVY_VERSION` (L36) is now unused. Since scanning is handled separately, these can be dropped to avoid the impression that this job feeds a scan.

.scripts/image-version.sh
```diff
@@ -0,0 +1,95 @@
...
+            > "$render_dir/$chart.yaml" 2> "$render_dir/$chart.err"; then
+        echo "::error::helm template for for '$chart':"
+        sed 's/^/    /' "$render_dir/$chart.err" >&2
+        FAILED+=("$chart:template"); continue
+    fi
+done
+
+IMAGES="$(
+  cat "$render_dir"/*.yaml |
+    grep -E '^[[:space:]]*image:' |
```
Matthiator: Coverage note (not a blocker): `grep image:` catches normal + init + sidecar containers (good!), but by design misses runtime-injected images (mutating webhooks, Kyverno, operator-set images via flags), image refs passed via `args`/`env`, and `- image:` list-item lines (the regex only matches `image:` after whitespace, not after `- `). It also only renders one values set, so disabled sub-features are missed. Might be worth a comment noting this is best-effort, so "collected" isn't read as "complete."

.scripts/image-version.sh
```diff
@@ -0,0 +1,95 @@
...
+    sed -E "s/^[[:space:]]*image:[[:space:]]*//; s/[\"']//g" |
+    grep -vE '[*!]' |          # drop kyverno wildcard/negation entries
+    grep -vE '^[[:space:]]*$' |
+    sort -u
+)"
+
+echo "Done Rendering!"
+
+[[ -n "$IMAGES" ]] || { echo "::warning::No image references found"; exit 0; }
```
Matthiator: With `set -euo pipefail`, if the `grep` above matches nothing the pipeline exits 1 and the script dies in the `IMAGES="$( … )"` assignment — before ever reaching this friendly `::warning::No image references found` / `exit 0`. A `|| true` on the grep would make the empty-case handling actually reachable.

.scripts/image-version.sh
```diff
@@ -0,0 +1,95 @@
...
+
+# pipefail that pipes break
+set -euo pipefail
+
+export PATH="$HOME/.local/bin/:$PATH"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+MANAGED="${MANAGED:-${PWD}/platform-components/helm}"
+CONFIG_FILE="${CONFIG_FILE:-config.yaml}"
+CLUSTER_NAME="$(yq -r '.clusters[0].name' "$CONFIG_FILE")"
```
Matthiator: Tiny thing: `yq` reads `$CONFIG_FILE` here before the `[[ -f … ]]` / `command -v` checks below, so a missing file or missing `yq` produces a cryptic command-substitution error instead of your nice diagnostic. Moving the checks above this line would surface the friendly message.

.scripts/kubara-config-update.sh
```diff
@@ -49,4 +49,14 @@ apply_yaml_if_set KUBARA_KUBERNETES_VERSION  ".clusters[0].terraform.kubernetesV
 apply_yaml_if_set KUBARA_DNS_NAME            ".clusters[0].dnsName"
 apply_yaml_if_set KUBARA_DNS_NAME            ".clusters[0].terraform.dns.name"
 
+# Enable every catalog service to the image/vuln report and the release header cover all charts
+yq eval '(.clusters[0].services[] | .status) = "enabled"' -i "$CFG"
+
+# metalb, loki and velero need custom configs
+yq eval '.clusters[0].services.metallb.config.publicLoadBalancerIPs = "203.0.113.10"' -i "$CFG"
+yq eval '.clusters[0].services.metallb.config.loadBalancerAddressPool = ["203.0.113.0/24"]' -i "$CFG"
+yq eval '.clusters[0].storage.bucketNames.chunks = "loki"' -i "$CFG"
```
Matthiator: This writes `.clusters[0].storage.bucketNames.chunks`, but `Cluster` has no top-level `storage` field, so mapstructure drops it before schema validation → silently ineffective. The intended Loki path is likely under `services.loki` / the generated values. Worth double-checking this actually has the intended effect.

.github/workflows/pr-checks.yaml
```diff
@@ -426,3 +428,52 @@ jobs:
...
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Checkout CI scripts
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          sparse-checkout: |
+            .scripts
+            .github/helm-profiles
```
Matthiator: Small cleanup: `.github/helm-profiles` is sparse-checked-out here but doesn't seem to be used anywhere in this job — can probably be dropped from the checkout.

tuunit:
I think we need to move and adapt this to the new repository

.github/workflows/pr-checks.yaml
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.github/workflows/release.yaml
davrad: This part will be rewritten as the image extraction now lives in another repo.

.scripts/image-version.sh
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.scripts/image-version.sh
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.scripts/image-version.sh
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.scripts/image-version.sh
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.scripts/image-version.sh
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.scripts/image-version.sh
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.scripts/kubara-config-update.sh
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

.github/workflows/pr-checks.yaml
davrad: Addressed in this PR [here](https://github.com/kubara-io/catalogs/pull/20).

davrad:
This PR has changed its scope. Instead of doing both image scanning for helm images, it will pull the extracted image versions, generated by another pipeline in the catalog repository and put it into the goreleaser pipeline. Other PR can be viewed [here](https://github.com/kubara-io/catalogs/pull/20).

.github/workflows/release.yaml
```diff
@@ -57,6 +63,49 @@ jobs:
...
+        id: images
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          set -euo pipefail
+
+          out="$RUNNER_TEMP/catalog-images"
+
+          for catalog in bootstrap general; do
+            oras pull "${CATALOG_IMAGES_REPO}/${catalog}-images:latest" -o "$out/$catalog"
```
Matthiator: After looking at the catalogs PR as well, I think the inventories belong to the catalog releases: bootstrap and general are versioned independently. Could we avoid pulling and copying the lists here, and instead link the catalog releases pinned by this Kubara version? That would also remove the GHCR dependency from the release job.

.github/workflows/release.yaml
```diff
@@ -57,6 +63,49 @@ jobs:
       - name: Install go-licenses
         run: go install github.com/google/go-licenses/v2@3e084b0caf710f7bfead967567539214f598c0a2 #v2.0.1
 
+      - name: Verify preinstalled Helm
```
Matthiator: Do we still need this step? Nothing in the release workflow uses Helm anymore, so I think we can remove it.

.github/workflows/release.yaml
```diff
@@ -57,6 +63,49 @@ jobs:
...
+      - name: Install oras
+        uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
+        with:
+          version: ${{ env.ORAS_VERSION }}
+
+
+      - name: Pull Images list from Catalog
+        id: images
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```
Matthiator: This token is a no-op here because ORAS reads the Docker credential store. Access to the new `*-images` packages isn’t proven yet either. If Kubara only links the catalog releases instead of pulling their inventories, this auth dependency goes away completely.

Matthiator (commented):
A few small suggestions to make the release inventory easier to follow and keep it tied to the actual Kubara release.

src/.goreleaser.yaml
```diff
@@ -13,6 +13,17 @@ release:
...
   extra_files:
     - glob: licenses.csv
+  footer: |
+    ## Platform images
+    Container images referenced by the platform components in this release
+    ```text
+    {{ .Env.KUBARA_IMAGE_LIST }}
+    ```
+    ## Helm Charts
+    Helm Chart Versions for Images:
```
Matthiator: I’d go one step further and keep the detailed inventories in the catalog releases. The Kubara release could just show its default catalog pins:

```yaml
## Default catalogs

- [Bootstrap catalog {{ .Env.BOOTSTRAP_CATALOG_VERSION }}](https://github.com/kubara-io/catalogs/releases/tag/bootstrap-v{{ .Env.BOOTSTRAP_CATALOG_VERSION }})
- [General catalog {{ .Env.GENERAL_CATALOG_VERSION }}](https://github.com/kubara-io/catalogs/releases/tag/general-v{{ .Env.GENERAL_CATALOG_VERSION }})

The actual images depend on the selected services and catalog overrides.
```

.gitignore
```diff
@@ -1,8 +1,9 @@
 .DS_Store
 .idea/
 .vscode/
+mise.toml
```
Matthiator: These changes look unrelated to the release inventory. Could we leave them out of this PR?

Matthiator (dismissed):
Thanks for putting this together! The image inventory is useful, but pulling mutable `latest` artifacts can make the release notes differ from the catalog versions actually pinned by Kubara. Once the workflow pulls the matching versioned catalog artifacts, I’m happy to take another look.

Matthiator (changes requested):
Thanks for putting this together! After looking at both PRs, I think the image inventories belong to the independently versioned catalog releases. Could we publish each catalog’s inventory with that catalog release, and have the Kubara release only link its pinned bootstrap and general releases? That keeps ownership clear and avoids making the Kubara release depend on mutable GHCR inventory artifacts.

.github/workflows/release.yaml
davrad: Updated: Link the release instead

.github/workflows/release.yaml
davrad: Removed it.

.github/workflows/release.yaml
davrad: Removed oras and auth token.

src/.goreleaser.yaml
davrad: Updated it.

Matthiator (approved):
Thanks for working through this! The release notes now link the catalog versions pinned by Kubara, and the workflow verifies that both releases exist. Looks good to me 🙂

PR #515, opened 2026-07-20
feat: declarative external catalogs

tuunit:
## 📝 Summary

This PR transitioning kubara from a single global embedded catalog to versioned, per-cluster catalog lists supported by a basic bootstrap catalog. This allows individual clusters to run different platform stacks while decoupling catalog versions from the CLI release cycle.

### Key changes:

- Layered Catalog Resolution: Catalogs are loaded sequentially
    - global bootstrap -> cluster-specific -> CLI flags
    - with automated OCI pulling and local caching
- Fixed Bootstrap Catalog: Necessary baseline for ArgoCD based stacks to work properly.
    - argo-cd and bootstrap-crds are therefore handled with a dedicated bootstrap catalog
    - Removing the need for separate --with-es-crds and --with-prometheus-crds bootstrap flags

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [x] 📦 Helm chart
- [x] 🧱 Terraform module
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [x] ♻️ Refactor / cleanup

## 🧪 Testing
- [x] CI passed
- [x] Unit tested
  
## ✅ Checklist


src/internal/render/render.go
```diff
@@ -126,55 +124,26 @@ func (tt TemplateType) String() string {
...
-		}
-		if !info.IsDir() {
-			return false, fmt.Errorf("%s template root %q is not a directory", source.name, root)
+			return nil, fmt.Errorf("resolve external catalog source: %w", err)
 		}
-		return true, nil
+		sources = append(sources, templateSource{
+			name:     cat,
+			fsys:     os.DirFS(source.RootPath),
+			baseRoot: ".",
```
Matthiator, line 140: `TemplateType=All` walks from this catalog root, so it also picks up `Catalog.yaml`. Since that path is not remapped to the configured work directory, it gets written to the process working directory. Is that intentional? If the manifest is only catalog metadata, could generation be limited to `platform-components` and `platform-configs`?

src/internal/render/render.go
```diff
@@ -126,55 +124,26 @@ func (tt TemplateType) String() string {
...
+			external: true,
```
Matthiator: All template sources are now marked as `external`, so the existing built-in/external comparison no longer distinguishes catalog layers. As a result, template collisions keep the first source and `--catalog-overwrite` cannot make the later catalog win. Could we preserve the source order or priority and use it for collision handling?

src/internal/config/store.go
```diff
@@ -286,7 +288,43 @@ func (cs *ConfigStore) GetCatalog() (catalog.Catalog, error) {
...
+		bootstrapCatalog = *cs.config.BootstrapCatalog
+	}
+
+	// Gather all unique catalogs from all clusters
+	allCatalogs := make([]string, 0)
+	seen := make(map[string]bool)
+
+	// Add cluster-specific catalogs in order
+	if cs.config != nil {
+		for _, cluster := range cs.config.Clusters {
```
Matthiator: `catalogs` is configured per cluster, but this combines the catalogs from all clusters into one global catalog. Defaults are then applied to every cluster, while generation loads only the current cluster templates. Could the effective catalog be resolved and applied separately for each cluster?

src/internal/config/factory.go
```diff
@@ -54,15 +57,28 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
...
 		ArgoCD:   argoCD,
+		Catalogs: append([]string(nil), effectiveCatalogOptions.Catalogs...),
 		Services: services,
 	}, nil
 }
 
+func clusterCatalogLoadOptions(catalogOptions catalog.LoadOptions) catalog.LoadOptions {
+	effective := catalogOptions
+	if len(effective.Catalogs) == 0 {
+		effective.Catalogs = []string{catalog.DefaultGeneralCatalog}
```
Matthiator, line 68: `kubara init` uses the default general catalog here, but `kubara schema` currently loads only the bootstrap catalog when no `--catalog` is provided. After bootstrap services are filtered, the generated schema contains no service definitions. Could both commands use the same default catalog resolution?

src/internal/config/factory.go
```diff
@@ -54,15 +57,28 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
...
+		return effective
+	}
+
+	effective.Catalogs = append([]string(nil), effective.Catalogs...)
```
Matthiator, line 72: An explicitly selected custom catalog replaces the default general catalog here. The bootstrap Argo CD template still directly indexes services such as `oauth2-proxy` and `kube-prometheus-stack`, so `generate` fails when those services are absent. Could the bootstrap templates use safe defaults and remain independent of the general catalog?

Matthiator:
Just a friendly reminder for the final steps before merging/releasing:

- fill in the PR description, including the breaking changes and catalog precedence
- update the affected catalog and bootstrap documentation
- get the remaining CI checks green
- make the `kubara-io/catalogs` repository and the `bootstrap` / `general` GHCR packages publicly readable before release

src/internal/render/render.go
tuunit: true missed it during testing because the TemplateAll doesn't get triggered if you do not have a terraform block defined. i removed this logic and ensured that we only render the platform-* directories and not the catalog root

src/internal/config/factory.go
tuunit: yes indeed the schema command used a "shortcut" and directly called the catalog instead of the config store which is wrong. therefore schema now relies on the existence of a config file and can generate schemas accordingly based on config files alone or in combination with additional catalog flags

src/internal/config/factory.go
tuunit: yep that needs to be fixed in the upstream catalogs repo

tuunit: fixed in https://github.com/kubara-io/catalogs/commit/1678fc95a6febedc780424436a0a584936d5be15

src/internal/render/render.go
tuunit: the unnecessary / legacy external field has been removed

src/internal/config/store.go
tuunit: fully refactored for per cluster catalog building and alignment of catalog loading between all commands

Matthiator (commented):
The rest of the changes look good to me. One small additional documentation leftover: `src/cmd/generate.go:38` still says “Renders embedded Helm and Terraform templates”.

src/cmd/schema.go
```diff
@@ -88,8 +97,12 @@ func (flags *SchemaFlags) AddFlags(cmd *cli.Command) {
 }
 
 func (o *SchemaOptions) Run() error {
-	// Generate schema
-	schemaDoc, err := config.GenerateSchemaWithCatalog(o.catalogOptions)
+	cs := config.NewConfigStore(o.cwd, o.configFilePath, o.catalogOptions)
+	if err := cs.Load(); err != nil && !errors.Is(err, os.ErrNotExist) {
```
Matthiator, line 101: Could we avoid persisting config migrations from `kubara schema`? `ConfigStore.Load()` saves migrated configurations, so generating a schema can unexpectedly rewrite `config.yaml`. Loading and migrating only in memory would keep this command side-effect free.

docs/content/4_building_your_platform/create_catalog.md
```diff
@@ -142,7 +156,7 @@ Typical reasons:
 - replace built-in templates with internal ones
 
 Without `--catalog-overwrite`, kubara rejects the collision.
```
Matthiator, line 158: Small documentation leftover: the nearby heading and examples still refer to “built-in services” and “built-in templates”, although these templates are now provided by catalogs. Could we update that wording as part of this PR?

src/cmd/schema.go
tuunit: I specifically do not want a different flow for the schema command. If an old config is used using a new version of the CLI. It should be migrated!

docs/content/4_building_your_platform/create_catalog.md
tuunit: Good point. Fixed and clarified as part of 49f6ca7

Matthiator (approved):
lgtm

PR #530, opened 2026-07-28
fix(catalog)!: accept password and identity token values

LSchu:
## 📝 Summary
The command line options `--password` and `--identity-token` for `kubara catalog login` were boolean flags, which didn't read the password/token as an argument as expected. Password/token values are now read from arguments and used for login.

This changes the CLI syntax: both flags now require a value. Interactive password login remains available by omitting `--password`; identity tokens can be provided through the flag or stdin.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📦 Helm chart
- [ ] 🧱 Terraform module
- [ ] 📝 Documentation
- [ ] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [x] CI passed
- [ ] Unit tested
  
## ✅ Checklist


src/cmd/catalog/login.go
```diff
@@ -112,6 +112,8 @@ func resolveCatalogLoginOptions(flags *catalogLoginFlags) (internal.LoginOptions
 		var identityToken string
 
 		switch {
+		case flags.identityToken != "":
```
Matthiator: Could we preserve the interactive identity-token flow here? `usesIdentityTokenAuth` is only true when a token value or `--identity-token-stdin` is set, so the `default` prompt below is now unreachable.

src/cmd/catalog/login.go
```diff
@@ -33,23 +33,23 @@ func NewCatalogLogin() *cli.Command {
 			&cli.StringFlag{
 				Name:        "username",
 				Aliases:     []string{"u"},
-				Usage:       "Log in using username and password",
+				Usage:       "registry username",
 				Destination: &flags.username,
 			},
-			&cli.BoolFlag{
+			&cli.StringFlag{
```
Matthiator, line 39: Could we update `catalog_distribution.md` as well? Its example uses `--password ghcr.io`, but this string flag will consume `ghcr.io` as the password and leave no registry argument. Removing `--password` from that interactive example should fix it.

src/cmd/catalog/login.go
```diff
@@ -33,23 +33,23 @@ func NewCatalogLogin() *cli.Command {
...
 				Name:        "password",
 				Aliases:     []string{"p"},
-				Usage:       "Log in with password interactively",
+				Usage:       "registry password",
```
Matthiator, line 42: Non-blocking: would it make sense to warn that passing a password directly may expose it in shell history or process output, and point users to the prompt or `--password-stdin`?

Matthiator (commented):
A few inline notes on the updated login flags.

Matthiator (changes requested):

Matthiator (commented):
One small docs detail is still left just below the changed section: the list says `identity token (interactive)`, but this PR removes that interactive prompt. Could you replace it with `identity token from flag`? GitHub won't let me add an apply-able suggestion there because that line isn't part of the diff.

docs/content/2_concepts/catalog_distribution.md
```diff
@@ -109,7 +109,13 @@ Only plain `major.minor.patch` is accepted.
...
 
 ```bash
-kubara catalog login -u my-github-user --password ghcr.io
+kubara catalog login -u my-github-user --password my-password ghcr.io
+```
+
+or alternatively
+
+```bash
+kubara catalog login -u my-github-user --password-stdin ghcr.io
```
Matthiator: Could we make the stdin flow explicit in the example? The current command works, but when pasted as-is it waits for terminal input without a prompt, which can look like it's stuck.

```suggestion
printf '%s' "$GITHUB_TOKEN" | kubara catalog login -u my-github-user --password-stdin ghcr.io
```

Matthiator (approved):
Looks good now, thanks!

PR #533, opened 2026-07-29
feat: renovate support for kubara catalogs

tuunit:
## 📝 Summary

* Adds a renovate config example to the docs
* Adds a automatically generated added renovate config when running `kubara init`

## 🧩 Type of change
- [X] 🔧 CLI / Go code
- [X] 📝 Documentation
- [X] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [X] CI passed
- [ ] Unit tested
  
## ✅ Checklist


Matthiator:
Great idea — generating `renovate.json` during `kubara init` makes automatic catalog updates much easier for our users.

I added a few follow-up adjustments to make it robust and release-ready:

- Generate `renovate.json` in the selected work directory and only when no Renovate configuration already exists, so we respect the user's project boundary and never overwrite an existing setup.
- Add `--renovate=false` as an opt-out for users who do not want Kubara to create the file.
- Use Renovate's current configuration fields and keep `config:recommended`, so the generated configuration validates against current Renovate versions and retains the normal Renovate behavior for other supported dependencies.
- Match catalog references from custom registries, including registries with ports, while excluding digest-pinned and tag-plus-digest references because those cannot be updated safely as regular versions.
- Write the file only after initialization succeeds, preventing a partial Renovate setup when `kubara init` fails.
- Detect Kubara's generated catalog manager in existing Renovate configurations, so users only get a warning when catalog updates are genuinely missing.
- Preserve characters such as `&` in the generated regex instead of JSON-encoding them as Unicode escapes, keeping the configuration readable.
- Add focused tests for the generation, matching, opt-out, existing-config, and failure scenarios, and document what the generated configuration does and how to extend an existing Renovate setup.

The original approach stays the same; these changes mainly cover compatibility and edge cases around it.

Matthiator (approved):

PR #540, opened 2026-08-04
fix(local): enable persistent openbao and resolve helm dependency lock bug

Rana-Ashar:
## Overview
This PR resolves the ephemeral OpenBao issue for local environments, ensuring that OpenBao state and secrets survive pod crashes and Docker reboots. Along the way, it also resolves two underlying bugs affecting the local bootstrap process.

**Resolves Ticket:** Improve local setup with persistent OpenBao

## What Changed
1. **Persistent OpenBao & Auto-Unseal Sidecar**
   - Disabled memory-only `dev` mode and enabled `standalone` mode backed by a 2Gi `file` PVC.
   - Injected a robust `auto-unsealer` sidecar container. The sidecar monitors the Vault API and automatically handles `bao operator init` (first boot) and `bao operator unseal` (subsequent boots/restarts) by storing and retrieving the keys from the persistent volume.
   - Refactored `kubectlExec` to dynamically extract the newly generated root token from the persistent volume so the CLI can securely configure auth and policies.

2. **Fixed `403 Permission Denied` for External Secrets**
   - **Bug:** The CLI was passing a static `token_reviewer_jwt` file path during OpenBao Kubernetes auth configuration. Modern Kubernetes dynamically rotates these tokens, causing `ExternalSecrets` to get locked out with a 403 error.
   - **Fix:** Removed the static token injection. OpenBao now natively uses the Kubernetes TokenReview API, allowing auth to survive pod restarts and token rotations perfectly.

3. **Fixed `helm dependency build` Lockfile Crash**
   - **Bug:** When rendering local charts, dynamically modifying the chart caused a mismatch between `Chart.yaml` and `Chart.lock` timestamps, crashing the Helm dependency step.
   - **Fix:** Changed the command in `dependency.go` from `"build"` to `"update"` to force Helm to regenerate the lockfile dynamically.

## Testing Performed
- [x] Ran `kubara bootstrap test-cluster --local` successfully.
- [x] Verified OpenBao initializes and unseals automatically on first boot.
- [x] Verified `ExternalSecrets` successfully binds to OpenBao (`ClusterSecretStore` is Ready).
- [x] Verified Grafana passwords dynamically sync into Kubernetes secrets.
- [x] **Disaster Recovery Test:** Deleted the `openbao-0` pod manually. Verified the sidecar automatically successfully unsealed the new pod  and no secrets were lost.

### DIFF
--- a/src/internal/cmd/bootstrap/local.go
+++ b/src/internal/cmd/bootstrap/local.go
@@ -42,6 +42,8 @@ const (
 	localExternalSecretsRole    = "any-sa"
 	localExternalSecretsPolicy  = "read-access"
 	localOpenBaoInternalAddress = "http://openbao.openbao.svc:8200"
+	// renovate: datasource=helm depName=openbao registryUrl=https://openbao.github.io/openbao-helm
+	localOpenBaoChartVersion = "0.29.0"
 )
 
 type localChart struct {
@@ -54,7 +56,7 @@ type localChart struct {
 }
 
 var (
-	localOpenBaoChart = localChart{"OpenBao", helm.RepoOptions{Name: "openbao", URL: "https://openbao.github.io/openbao-helm"}, localOpenBaoReleaseName, "openbao/openbao", "0.28.3", localOpenBaoNamespace}
+	localOpenBaoChart = localChart{"OpenBao", helm.RepoOptions{Name: "openbao", URL: "https://openbao.github.io/openbao-helm"}, localOpenBaoReleaseName, "openbao/openbao", localOpenBaoChartVersion, localOpenBaoNamespace}
 )
 
 func prepareLocalBootstrap(ctx context.Context, opts *Options) error {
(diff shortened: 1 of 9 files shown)


Matthiator:
Upgrade note for existing local development clusters: clusters created before this PR used OpenBao dev mode without persistent storage. Because Kubernetes does not allow adding `volumeClaimTemplates` to an existing StatefulSet, these clusters should be recreated once:

```bash
kind delete cluster --name <cluster-name>
kubara bootstrap --local <cluster-name>
```

No automatic migration is included because the previous local setup did not persist OpenBao data.

Matthiator (dismissed):
Thanks for the great work on this! 🙌

I successfully tested the complete local bootstrap with a fresh kind cluster. OpenBao initialized and auto-unsealed correctly after both recreating the pod and restarting the complete kind node. The PVC, initialization data, and a test KV secret remained intact.

I also added a few follow-up improvements: Renovate can now detect the pinned OpenBao chart version, and the sidecar inherits the chart-managed OpenBao image so both containers stay in sync. I also fixed the ingress TLS values structure, hardened shell and token handling, preserved wrapped errors while redacting the root token, and improved Helm dependency handling for missing or outdated lock files. The related tests and documentation were updated as well.

Matthiator (approved):

src/internal/cmd/bootstrap/local.go
```diff
@@ -512,9 +516,7 @@ func configureLocalOpenBao(ctx context.Context, kubeconfigPath string, cluster *
 		"exec", localOpenBaoPodName,
 		"--",
 		"bao", "write", "auth/kubernetes/config",
-		"token_reviewer_jwt=@/var/run/secrets/kubernetes.io/serviceaccount/token",
 		"kubernetes_host=https://kubernetes.default.svc:443",
-		"kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
```
tuunit, line 517: why was this removed?

Matthiator: I removed both values because OpenBao recommends omitting token_reviewer_jwt and kubernetes_ca_cert when it runs in a Kubernetes pod. It then uses the locally mounted service account token and CA certificate and periodically re-reads the token to support short-lived tokens.
This is documented here: https://openbao.org/docs/auth/kubernetes/#use-local-service-account-token-as-the-reviewer-jwt

PR #572, opened 2026-08-17
feat(bootstrap): provision STACKIT secret store

Matthiator:
## 📝 Summary

Automatically provision the STACKIT Secrets Manager credential Secret and `ClusterSecretStore` during bootstrap when `external-secrets` is enabled.

The bootstrap flow now:

- respects `--with-es-css-file` as an explicit override
- reuses an existing `<cluster>-<stage>` store
- reads the STACKIT read-only username, password, API URL, and mount path from Terraform or OpenTofu outputs when no store exists
- applies and verifies both the credential Secret and ClusterSecretStore
- skips all discovery and provisioning when `external-secrets` is disabled
- retains the generated T Cloud Public OpenBao flow

Users of another secret backend can continue to provide a manifest with `--with-es-css-file` or create the expected store before bootstrap.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested

Additional validation:

- `go test ./...`
- `go vet ./...`
- `make build-binary`
- `make docs-validate`
- generated a STACKIT SKE platform from the working-tree catalog
- verified the rendered Terraform outputs with `terraform fmt -check`

## 🔗 Additional Context / Related Issues / Tickets

Requires kubara-io/catalogs#72 for the new `vault_api_url` and `vault_path` outputs.

The automatic path runs only for STACKIT. Custom secret backends remain supported through an existing store or `--with-es-css-file`.

## ✅ Checklist

### DIFF
--- a/docs/content/1_getting_started/bootstrapping.md
+++ b/docs/content/1_getting_started/bootstrapping.md
@@ -273,9 +273,15 @@ CI-specific values can be stored in chart-local CI files (for example `ci/ci-val
 
 External Secrets needs a `ClusterSecretStore`.
 
+If the `external-secrets` service is disabled, kubara skips all store discovery and automatic setup.
+
 For T Cloud Public CCE, kubara already renders the OpenBao-backed `ClusterSecretStore` into `external-secrets/values.yaml`. Do not pass `--with-es-css-file` and do not create a separate provider credential secret.
 
-For other providers, create the provider credential secret first and either pass a `ClusterSecretStore` manifest during bootstrap with `--with-es-css-file`, or apply the `ClusterSecretStore` manually after bootstrap.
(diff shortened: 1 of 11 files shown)


tuunit:
This is pure feature creep. No abstraction for IaC handling has been introduced and the contents and task of the packages aren't properly separated. I don't think we should continue on this tangent and instead re-evaluate on a larger scale how we want to deal with infrastructure in the future. As our focus is the platform on top of kubernetes and less the infra part we as a team need to discuss how to move this forward. This PR would introduce too many hardcoded assumptions and dependencies

PR #599, opened 2026-09-03
feat(bootstrap): add spoke credential onboarding

Rana-Ashar:
## 📝 Summary

Adds spoke credential onboarding support to `kubara bootstrap`.

A spoke can now be onboarded using an initial kubeconfig:

```bash
kubara bootstrap <spoke-name> \
  --initial-kubeconfig <path> \
  --namespace <agent-namespace>

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [ ] 📝 Documentation
- [ ] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [ ] Unit tested
  
## ✅ Checklist

### DIFF
--- a/src/cmd/bootstrap.go
+++ b/src/cmd/bootstrap.go
@@ -29,6 +29,9 @@ type BootstrapFlags struct {
 	EnvPrefixFlag          string
 	DryRun                 bool
 	Timeout                time.Duration
+
+	InitialKubeconfig string
+	AgentNamespace    string
 }
 
 var deprecatedBootstrapCRDFlags = []string{
@@ -38,9 +41,10 @@ var deprecatedBootstrapCRDFlags = []string{
 
 func NewBootstrapFlags() *BootstrapFlags {
 	return &BootstrapFlags{
-		EnvFile:       ".env",
-		EnvPrefixFlag: "KUBARA_",
-		Timeout:       2 * time.Minute,
+		EnvFile:        ".env",
+		EnvPrefixFlag:  "KUBARA_",
+		Timeout:        2 * time.Minute,
+		AgentNamespace: "kubara",
 	}
 }
 
@@ -48,11 +52,12 @@ func NewBootstrapCmd() *cli.Command {
 	flags := NewBootstrapFlags()
 
 	cmd := &cli.Command{
-		Name:        "bootstrap",
-		Usage:       "Bootstrap Argo CD onto a cluster",
-		UsageText:   "kubara bootstrap CLUSTER_NAME [--local]",
+		Name:  "bootstrap",
+		Usage: "Bootstrap a hub or onboard a spoke cluster",
+		UsageText: `kubara bootstrap CLUSTER_NAME [--local]
+kubara bootstrap CLUSTER_NAME --initial-kubeconfig PATH [--namespace NAMESPACE]`,
 		ArgsUsage:   "CLUSTER_NAME",
-		Description: "Bootstraps Argo CD onto the specified cluster. The optional --local mode provisions an isolated local evaluation environment and is not intended for production use.",
+		Description: "Bootstraps Argo CD onto a hub cluster, or onboards a spoke cluster when --initial-kubeconfig is provided.",
 		Arguments: []cli.Argument{
 			&cli.StringArg{
 				Name:      "cluster-name",
@@ -62,19 +67,22 @@ func NewBootstrapCmd() *cli.Command {
 		Action: func(c context.Context, cmd *cli.Command) error {
 			printDeprecatedBootstrapCRDFlagNotice(cmd)
 
+			clusterName := cmd.StringArg("cluster-name")
+			if clusterName == "" {
+				return fmt.Errorf("missing argument %q", "cluster-name")
+			}
+
 			o, err := flags.ToOptions(cmd)
 			if err != nil {
 				return fmt.Errorf("convert flags to options: %w", err)
 			}
-			if cmd.StringArg("cluster-name") == "" {
-				return fmt.Errorf("missing argument %q", "cluster-name")
-			}
-			o.ClusterName = cmd.StringArg("cluster-name")
+
+			o.ClusterName = clusterName
 			return Run(c, o)
 		},
 	}
-	flags.AddFlags(cmd)
 
+	flags.AddFlags(cmd)
 	return cmd
 }
 
@@ -84,14 +92,115 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
 		return nil, fmt.Errorf("get working directory: %w", err)
 	}
 
+	kubeconfigPath := cmd.String("kubeconfig")
+	if !cmd.IsSet("kubeconfig") {
+		if envKubeconfig := os.Getenv("KUBECONFIG"); envKubeconfig != "" {
+			kubeconfigPath = envKubeconfig
+		}
+	}
+
+	kubeconf, err := utils.GetFullPath(kubeconfigPath, cwd)
+	if err != nil {
+		return nil, fmt.Errorf("get kubeconfig path: %w", err)
+	}
+
+	catalogOptions, err := catalogLoadOptionsFromCommand(cmd, "")
+	if err != nil {
+		return nil, fmt.Errorf("get catalog options: %w", err)
+	}
+
+	configFilePath, err := utils.GetFullPath(cmd.String("config-file"), cwd)
+	if err != nil {
+		return nil, fmt.Errorf("get config file path: %w", err)
+	}
+
+	cs := config.NewConfigStore(cwd, configFilePath, catalogOptions)
+	if err := cs.Load(); err != nil {
+		return nil, fmt.Errorf("load config: %w", err)
+	}
+
+	clusterName := cmd.StringArg("cluster-name")
+
+	var clusterConfig *config.Cluster
+	for i := range cs.GetConfig().Clusters {
+		if cs.GetConfig().Clusters[i].Name == clusterName {
+			clusterConfig = &cs.GetConfig().Clusters[i]
+			break
+		}
+	}
+
+	if clusterConfig == nil {
+		return nil, fmt.Errorf(
+			"cluster %q not found in config file %q",
+			clusterName,
+			configFilePath,
+		)
+	}
+
+	// Spoke onboarding is intentionally routed before the existing hub bootstrap
+	// environment/catalog preparation. The spoke path only needs the Kubara
+	// configuration, a Hub kubeconfig, and the initial Spoke kubeconfig.
+	if flags.InitialKubeconfig != "" {
+		if flags.Local {
+			return nil, fmt.Errorf(
+				"--local cannot be used with --initial-kubeconfig",
+			)
+		}
+
+		if flags.DryRun {
+			return nil, fmt.Errorf(
+				"--dry-run is not supported with --initial-kubeconfig",
+			)
+		}
+
+		if clusterConfig.Type != "spoke" {
+			return nil, fmt.Errorf(
+				"--initial-kubeconfig can only be used with a spoke cluster; cluster %q has type %q",
+				clusterConfig.Name,
+				clusterConfig.Type,
+			)
+		}
+
+		initialKubeconfig, err := utils.GetFullPath(
+			flags.InitialKubeconfig,
+			cwd,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"resolve initial kubeconfig path: %w",
+				err,
+			)
+		}
+
+		if _, err := os.Stat(initialKubeconfig); err != nil {
+			return nil, fmt.Errorf(
+				"initial kubeconfig %q is not accessible: %w",
+				initialKubeconfig,
+				err,
+			)
+		}
+
+		return &bootstrap.Options{
+			Kubeconfig:        kubeconf,
+			InitialKubeconfig: initialKubeconfig,
+			AgentNamespace:    flags.AgentNamespace,
+			ClusterConfig:     clusterConfig,
+			ClusterName:       clusterName,
+			Timeout:           flags.Timeout,
+			WorkDir:           cwd,
+			ConfigFilePath:    configFilePath,
+		}, nil
+	}
+
+	if cmd.IsSet("namespace") {
+		return nil, fmt.Errorf(
+			"--namespace can only be used together with --initial-kubeconfig",
+		)
+	}
+
 	envFilePath, err := utils.GetFullPath(cmd.String("env-file"), cwd)
 	if err != nil {
 		return nil, fmt.Errorf("get env file path: %w", err)
-	}
-
-	kubeconf, err := utils.GetFullPath(cmd.String("kubeconfig"), cwd)
-	if err != nil {
-		return nil, fmt.Errorf("get kubeconfig path: %w", err)
 	}
 
 	componentsAbsPath := flags.PlatformComponentsPath
@@ -112,35 +221,6 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
 		}
 	}
 
-	catalogOptions, err := catalogLoadOptionsFromCommand(cmd, "")
-	if err != nil {
-		return nil, fmt.Errorf("get catalog options: %w", err)
-	}
-
-	// Load config file and find cluster by name
-	configFilePath, err := utils.GetFullPath(cmd.String("config-file"), cwd)
-	if err != nil {
-		return nil, fmt.Errorf("get config file path: %w", err)
-	}
-
-	cs := config.NewConfigStore(cwd, configFilePath, catalogOptions)
-	if err := cs.Load(); err != nil {
-		return nil, fmt.Errorf("load config: %w", err)
-	}
-
-	// Find the cluster by name from the argument
-	clusterName := cmd.StringArg("cluster-name")
-	var clusterConfig *config.Cluster
-	for i := range cs.GetConfig().Clusters {
-		if cs.GetConfig().Clusters[i].Name == clusterName {
-			clusterConfig = &cs.GetConfig().Clusters[i]
-			break
-		}
-	}
-	if clusterConfig == nil {
-		return nil, fmt.Errorf("cluster %q not found in config file %q", clusterName, configFilePath)
-	}
-
 	es := envconfig.NewEnvStore(envFilePath, ".", flags.EnvPrefixFlag)
 	if err := es.Load(); err != nil {
 		return nil, fmt.Errorf("load env: %w", err)
@@ -151,28 +231,36 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
 		return nil, fmt.Errorf("prepare env: %w", err)
 	}
 
-	// Validate and normalize ClusterSecretStore path if provided
 	var cssAbsPath string
 	if flags.ClusterSecretStorePath != "" {
 		if !filepath.IsAbs(flags.ClusterSecretStorePath) {
 			cssAbsPath = filepath.Join(cwd, flags.ClusterSecretStorePath)
 			cssAbsPath, err = filepath.Abs(cssAbsPath)
 			if err != nil {
-				return nil, fmt.Errorf("getting absolute path for ClusterSecretStore file: %w", err)
+				return nil, fmt.Errorf(
+					"getting absolute path for ClusterSecretStore file: %w",
+					err,
+				)
 			}
 		} else {
 			cssAbsPath = flags.ClusterSecretStorePath
 		}
 
-		// Verify file exists
 		if _, err := os.Stat(cssAbsPath); err != nil {
-			return nil, fmt.Errorf("cluster secret store file not found: %w", err)
+			return nil, fmt.Errorf(
+				"cluster secret store file not found: %w",
+				err,
+			)
 		}
 	}
 
 	loadedCatalog, err := cs.GetCatalogForCluster(*clusterConfig)
 	if err != nil {
-		return nil, fmt.Errorf("load catalog for cluster %q: %w", clusterConfig.Name, err)
+		return nil, fmt.Errorf(
+			"load catalog for cluster %q: %w",
+			clusterConfig.Name,
+			err,
+		)
 	}
 
 	bootstrapCatalog := catalog.DefaultBootstrapCatalog
@@ -256,13 +344,34 @@ func (flags *BootstrapFlags) AddFlags(cmd *cli.Command) {
 			Usage:       "Timeout for kubernetes API calls (e.g. 10s, 1m)",
 			Destination: &flags.Timeout,
 		},
+		&cli.StringFlag{
+			Name:        "initial-kubeconfig",
+			Usage:       "Path to the initial kubeconfig used to onboard a spoke cluster",
+			Destination: &flags.InitialKubeconfig,
+			Config: cli.StringConfig{
+				TrimSpace: true,
+			},
+		},
+		&cli.StringFlag{
+			Name:        "namespace",
+			Value:       flags.AgentNamespace,
+			Usage:       "Namespace for the kubara-agent ServiceAccount on the spoke cluster",
+			Destination: &flags.AgentNamespace,
+			Config: cli.StringConfig{
+				TrimSpace: true,
+			},
+		},
 	}
 
 	cmd.Flags = append(cmd.Flags, bootstrapFlags...)
 }
 
 func Run(ctx context.Context, o *bootstrap.Options) error {
-	ctx, cancelSignal := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
+	ctx, cancelSignal := signal.NotifyContext(
+		ctx,
+		os.Interrupt,
+		syscall.SIGTERM,
+	)
 	defer cancelSignal()
 
 	return bootstrap.Bootstrap(ctx, o)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/cmd/bootstrap.go [hunk 1]
```diff
@@ -29,6 +29,9 @@ type BootstrapFlags struct {
 	EnvPrefixFlag          string
 	DryRun                 bool
 	Timeout                time.Duration
+
+	InitialKubeconfig string
+	AgentNamespace    string
```
tuunit: remove empty line, this will automatically align the types when formatting

src/cmd/bootstrap.go
```diff
@@ -38,21 +41,23 @@ var deprecatedBootstrapCRDFlags = []string{
...
-		Name:        "bootstrap",
-		Usage:       "Bootstrap Argo CD onto a cluster",
-		UsageText:   "kubara bootstrap CLUSTER_NAME [--local]",
+		Name:  "bootstrap",
+		Usage: "Bootstrap a hub or onboard a spoke cluster",
+		UsageText: `kubara bootstrap CLUSTER_NAME [--local]
+kubara bootstrap CLUSTER_NAME --initial-kubeconfig PATH [--namespace NAMESPACE]`,
 		ArgsUsage:   "CLUSTER_NAME",
-		Description: "Bootstraps Argo CD onto the specified cluster. The optional --local mode provisions an isolated local evaluation environment and is not intended for production use.",
+		Description: "Bootstraps Argo CD onto a hub cluster, or onboards a spoke cluster when --initial-kubeconfig is provided.",
```
tuunit:
```suggestion
		Description: "Bootstraps Argo CD onto a hub cluster, or onboards a spoke cluster. When the option --local is defined it automatically provisions a isolated local environment using kind.",
```

tuunit: what happens if you define `--local` and `--initial-kubeconfig` ?

src/cmd/bootstrap.go
```diff
@@ -62,19 +67,22 @@ func NewBootstrapCmd() *cli.Command {
...
 			o, err := flags.ToOptions(cmd)
 			if err != nil {
 				return fmt.Errorf("convert flags to options: %w", err)
 			}
-			if cmd.StringArg("cluster-name") == "" {
-				return fmt.Errorf("missing argument %q", "cluster-name")
-			}
-			o.ClusterName = cmd.StringArg("cluster-name")
+
+			o.ClusterName = clusterName
```
tuunit, line 79: why change the order of operations? why first parse the cluster name?

tuunit: i mean makes sense to some extend to parse the cluster-name first as it is the most important value 🤔

src/cmd/bootstrap.go
```diff
@@ -84,40 +92,23 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
 	}
 
-	envFilePath, err := utils.GetFullPath(cmd.String("env-file"), cwd)
-	if err != nil {
-		return nil, fmt.Errorf("get env file path: %w", err)
+	kubeconfigPath := cmd.String("kubeconfig")
+	if !cmd.IsSet("kubeconfig") {
+		if envKubeconfig := os.Getenv("KUBECONFIG"); envKubeconfig != "" {
+			kubeconfigPath = envKubeconfig
+		}
```
tuunit, line 98:
```suggestion
		kubeconfigPath = os.Getenv("KUBECONFIG")
```

wouldn't this be equivalent and therefore easier?

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+			"cluster %q not found in config file %q",
+			clusterName,
+			configFilePath,
+		)
+	}
+
+	// Spoke onboarding is intentionally routed before the existing hub bootstrap
+	// environment/catalog preparation. The spoke path only needs the Kubara
+	// configuration, a Hub kubeconfig, and the initial Spoke kubeconfig.
+	if flags.InitialKubeconfig != "" {
```
tuunit: we do have the cluster config at this point so we should check by type and not for this flag

```go
	if clusterConfig.Type == config.Spoke {
		return bootstrapSpoke(...)
	}

  return bootstrapHub(...)
```

tuunit: refactor this function into two flows / methods

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+			)
+		}
+
+		if flags.DryRun {
+			return nil, fmt.Errorf(
+				"--dry-run is not supported with --initial-kubeconfig",
+			)
+		}
+
+		if clusterConfig.Type != "spoke" {
```
tuunit: don't use "magic" strings. use the constant from the config package

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+		)
+	}
+
+	// Spoke onboarding is intentionally routed before the existing hub bootstrap
+	// environment/catalog preparation. The spoke path only needs the Kubara
+	// configuration, a Hub kubeconfig, and the initial Spoke kubeconfig.
+	if flags.InitialKubeconfig != "" {
+		if flags.Local {
+			return nil, fmt.Errorf(
+				"--local cannot be used with --initial-kubeconfig",
```
tuunit: why not? would it make a difference anyway? 🤔

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+			)
+		}
+
+		if flags.DryRun {
+			return nil, fmt.Errorf(
+				"--dry-run is not supported with --initial-kubeconfig",
```
tuunit:
```suggestion
				"--dry-run is not supported for spoke clusters",
```

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+			)
+		}
+
+		if clusterConfig.Type != "spoke" {
+			return nil, fmt.Errorf(
+				"--initial-kubeconfig can only be used with a spoke cluster; cluster %q has type %q",
```
tuunit: the comments all focus on the fact that the `--initial-kubeconfig` flag was used but that is like diagnosing a symptom instead of the root cause and the root cause is the fact that we are in the flow for onboarding a spoke

tuunit: if the intitial check at the top already checks for the type. this check is obsolete anyways. so please invert the check

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+
+		initialKubeconfig, err := utils.GetFullPath(
+			flags.InitialKubeconfig,
+			cwd,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"resolve initial kubeconfig path: %w",
+				err,
+			)
```
tuunit:
```suggestion
			return nil, fmt.Errorf("resolve initial kubeconfig path: %w", err)
```

its still a short line therefore please don't introduce unnecessary linebreaks. makes it harder to read/scroll through the code

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+		}
+
+		if _, err := os.Stat(initialKubeconfig); err != nil {
+			return nil, fmt.Errorf(
+				"initial kubeconfig %q is not accessible: %w",
+				initialKubeconfig,
+				err,
+			)
```
tuunit:
```suggestion
			return nil, fmt.Errorf(
				"initial kubeconfig %q is not accessible: %w",
				initialKubeconfig, err,
			)
```

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+			InitialKubeconfig: initialKubeconfig,
+			AgentNamespace:    flags.AgentNamespace,
+			ClusterConfig:     clusterConfig,
+			ClusterName:       clusterName,
+			Timeout:           flags.Timeout,
+			WorkDir:           cwd,
+			ConfigFilePath:    configFilePath,
+		}, nil
+	}
+
```
tuunit, line 177:
```suggestion
	return bootstrapHub(...)
```

src/cmd/bootstrap.go
```diff
@@ -128,17 +119,106 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
+	if cmd.IsSet("namespace") {
+		return nil, fmt.Errorf(
+			"--namespace can only be used together with --initial-kubeconfig",
+		)
+	}
```
tuunit: spoke

src/cmd/bootstrap.go
```diff
@@ -151,28 +231,36 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
 	if flags.ClusterSecretStorePath != "" {
 		if !filepath.IsAbs(flags.ClusterSecretStorePath) {
 			cssAbsPath = filepath.Join(cwd, flags.ClusterSecretStorePath)
 			cssAbsPath, err = filepath.Abs(cssAbsPath)
 			if err != nil {
-				return nil, fmt.Errorf("getting absolute path for ClusterSecretStore file: %w", err)
+				return nil, fmt.Errorf(
+					"getting absolute path for ClusterSecretStore file: %w",
+					err,
+				)
```
tuunit: revert

src/cmd/bootstrap.go
```diff
@@ -151,28 +231,36 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
 			cssAbsPath = flags.ClusterSecretStorePath
 		}
 
-		// Verify file exists
 		if _, err := os.Stat(cssAbsPath); err != nil {
-			return nil, fmt.Errorf("cluster secret store file not found: %w", err)
+			return nil, fmt.Errorf(
+				"cluster secret store file not found: %w",
+				err,
+			)
```
tuunit: revert

src/cmd/bootstrap.go
```diff
@@ -151,28 +231,36 @@ func (flags *BootstrapFlags) ToOptions(cmd *cli.Command) (*bootstrap.Options, er
...
 	}
 
 	loadedCatalog, err := cs.GetCatalogForCluster(*clusterConfig)
 	if err != nil {
-		return nil, fmt.Errorf("load catalog for cluster %q: %w", clusterConfig.Name, err)
+		return nil, fmt.Errorf(
+			"load catalog for cluster %q: %w",
+			clusterConfig.Name,
+			err,
+		)
```
tuunit: revert

src/cmd/bootstrap.go
```diff
@@ -256,13 +344,34 @@ func (flags *BootstrapFlags) AddFlags(cmd *cli.Command) {
...
 	cmd.Flags = append(cmd.Flags, bootstrapFlags...)
 }
 
 func Run(ctx context.Context, o *bootstrap.Options) error {
-	ctx, cancelSignal := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
+	ctx, cancelSignal := signal.NotifyContext(
+		ctx,
+		os.Interrupt,
+		syscall.SIGTERM,
+	)
```
tuunit: revert

src/cmd/bootstrap.go
```diff
@@ -271,29 +380,43 @@ func Run(ctx context.Context, o *bootstrap.Options) error {
 func printDeprecatedBootstrapCRDFlagNotice(cmd *cli.Command) {
 	for _, flagName := range deprecatedBootstrapCRDFlags {
 		if cmd.IsSet(flagName) {
-			log.Warn().Msgf("--%s is deprecated and ignored; CRDs are applied automatically during bootstrap.\n", flagName)
+			log.Warn().
+				Msgf(
+					"--%s is deprecated and ignored; CRDs are applied automatically during bootstrap.\n",
+					flagName,
+				)
```
tuunit: revert

src/cmd/bootstrap.go
```diff
@@ -271,29 +380,43 @@ func Run(ctx context.Context, o *bootstrap.Options) error {
...
-	if (envconfig.IsConfiguredEnvValue(envMap.ArgocdGitUsername) && !envconfig.IsConfiguredEnvValue(envMap.ArgocdGitPatOrPassword)) ||
-		(!envconfig.IsConfiguredEnvValue(envMap.ArgocdGitUsername) && envconfig.IsConfiguredEnvValue(envMap.ArgocdGitPatOrPassword)) {
-		return nil, fmt.Errorf("if you are using a private repository you need to configure both ARGOCD_GIT_PAT_OR_PASSWORD and ARGOCD_GIT_USERNAME")
+	if (envconfig.IsConfiguredEnvValue(envMap.ArgocdGitUsername) &&
+		!envconfig.IsConfiguredEnvValue(envMap.ArgocdGitPatOrPassword)) ||
+		(!envconfig.IsConfiguredEnvValue(envMap.ArgocdGitUsername) &&
+			envconfig.IsConfiguredEnvValue(envMap.ArgocdGitPatOrPassword)) {
+		return nil, fmt.Errorf(
+			"if you are using a private repository you need to configure both ARGOCD_GIT_PAT_OR_PASSWORD and ARGOCD_GIT_USERNAME",
+		)
```
tuunit: revert

src/cmd/bootstrap.go
```diff
@@ -271,29 +380,43 @@ func Run(ctx context.Context, o *bootstrap.Options) error {
...
 	}
 
 	prepared := *envMap
+
 	if !envconfig.IsConfiguredEnvValue(prepared.ProjectName) {
 		prepared.ProjectName = cluster.Name
 	}
+
 	if !envconfig.IsConfiguredEnvValue(prepared.ProjectStage) {
 		prepared.ProjectStage = cluster.Stage
```
tuunit, line 384: revert

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	spokeBootstrapHubNamespace = "argocd"
```
tuunit: this name is way too convoluted and not precise. its not just a namespace for bootstrapping but where the access credentials will live in the future as well


```suggestion
	spokeCredentialsNamespace = "argocd"
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+
+	rotatorLabelKey   = "kubara.io/cluster-credential-rotator"
+	rotatorLabelValue = "true"
```
tuunit: why does this label exist?

didn't we decide to use / select by the secret type `kubara.io/argocd-cluster-bootstrap`?

tuunit: some for this type we should refactor it similar to what we already did for the service account. make it more generic:

`kubara.io/cluster-credentials`

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+
+	agentNamespaceAnnotation = "kubara.io/agent-namespace"
+
+	bootstrapKubeconfigKey = "bootstrapKubeconfig"
+	generatedConfigKey     = "config"
+
+	bootstrapSecretType = "kubara.io/argocd-cluster-bootstrap"
+
+	clusterConfigMapSuffix = "-config"
+	clusterSecretSuffix    = "-cluster-secret"
```
tuunit: unnecessary constants. remove and use in-place instead

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+
+	argoClusterLabel      = "argocd.argoproj.io/secret-type"
+	argoClusterLabelValue = "cluster"
+
+	rotatorSelector      = "app.kubernetes.io/name=cluster-credential-rotator"
```
tuunit: unnecessary constant: use in-place

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+	rotatorContainerName = "cluster-credential-rotator"
```
tuunit: the rotator only has one container. so why even have this?

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+			)
+		}
+	}
+
+	if !foundContainer {
+		return nil, fmt.Errorf(
+			"credential rotator CronJob does not contain container %q",
+			rotatorContainerName,
+		)
+	}
```
tuunit: what is this even for??


```suggestion
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+		Create(
+			ctx,
+			job,
+			metav1.CreateOptions{},
+		)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"create credential rotator Job: %w",
+			err,
+		)
```
tuunit:
```suggestion
		return nil, fmt.Errorf("create credential rotator Job: %w", err)
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+	}
+
+	createdJob, err := client.BatchV1().
+		Jobs(spokeBootstrapHubNamespace).
+		Create(
+			ctx,
+			job,
+			metav1.CreateOptions{},
+		)
```
tuunit:
```suggestion
		Create(ctx, job, metav1.CreateOptions{})
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+
+		return nil, fmt.Errorf(
+			"found multiple credential rotator CronJobs in namespace %q: %s",
+			spokeBootstrapHubNamespace,
+			strings.Join(names, ", "),
+		)
+	}
+}
+
+func createTargetedRotatorJob(
```
tuunit:
```suggestion
func startOnboardingJob(
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+	}
+
+	log.Info().Msgf("Updated initial cluster secret %q", secretName)
+	return nil
+}
+
+func findCredentialRotatorCronJob(
+	ctx context.Context,
+	client kubernetes.Interface,
+) (*batchv1.CronJob, error) {
```
tuunit: is this really "finding". its a basic client getter therefore we should call it that

```suggestion
func getCredsRotatorCronJob(ctx context.Context, client kubernetes.Interface) (*batchv1.CronJob, error) {
```

tuunit: why is the return a pointer if the list doesn't give you a pointer?

please use the type directly not a pointer

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+			ctx,
+			metav1.ListOptions{
+				LabelSelector: rotatorSelector,
+			},
+		)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"discover credential rotator CronJob: %w",
+			err,
+		)
```
tuunit:
```suggestion
		return nil, fmt.Errorf("discover credential rotator CronJob: %w",	err)
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+	client kubernetes.Interface,
+) (*batchv1.CronJob, error) {
+	cronJobs, err := client.BatchV1().
+		CronJobs(spokeBootstrapHubNamespace).
+		List(
+			ctx,
+			metav1.ListOptions{
+				LabelSelector: rotatorSelector,
+			},
+		)
```
tuunit:
```suggestion
		List(ctx, metav1.ListOptions{LabelSelector: rotatorSelector})
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+		for _, item := range cronJobs.Items {
+			names = append(names, item.Name)
+		}
+
+		return nil, fmt.Errorf(
+			"found multiple credential rotator CronJobs in namespace %q: %s",
+			spokeBootstrapHubNamespace,
+			strings.Join(names, ", "),
+		)
+	}
```
tuunit:
```suggestion
	if len(cronJobs.Items) == 1 {
		return &cronJobs.Items[0], nil
	}
	
	if len(cronJobs.items) > 1 {
		return nil, fmt.Errorf(
			"invalid setup: found multiple credential rotator CronJobs in namespace %q",
			spokeBootstrapHubNamespace,
		)
	}
	
	return nil, fmt.Errorf(
			"credential rotator CronJob not found in namespace %q; ensure the Hub has reconciled the credential rotator",
			spokeBootstrapHubNamespace,
	)
```

tuunit: early return and reduced nesting

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+
+	return createdJob, nil
+}
+
+func waitForRotatorJob(
+	ctx context.Context,
+	client kubernetes.Interface,
+	jobName string,
+) error {
```
tuunit:
```suggestion
func waitForRotatorJob(ctx context.Context,	client kubernetes.Interface, jobName string) error {
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+	defer ticker.Stop()
+
+	for {
+		job, err := client.BatchV1().
+			Jobs(spokeBootstrapHubNamespace).
+			Get(
+				ctx,
+				jobName,
+				metav1.GetOptions{},
+			)
```
tuunit:
```suggestion
			Get(ctx, jobName,	metav1.GetOptions{})
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+		if err != nil {
+			return fmt.Errorf(
+				"read credential rotator Job %q: %w",
+				jobName,
+				err,
+			)
```
tuunit:
```suggestion
			return fmt.Errorf("read credential rotator Job %q: %w", jobName, err)
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+
+func waitForRotatorJob(
+	ctx context.Context,
+	client kubernetes.Interface,
+	jobName string,
+) error {
+	ticker := time.NewTicker(2 * time.Second)
+	defer ticker.Stop()
+
+	for {
```
tuunit, line 356: this is potentially an endless loop. we need a timeout, lets say 5minutes

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+				}
+				if message == "" {
+					message = "Job reported failed condition"
+				}
+
+				return fmt.Errorf(
+					"credential rotator Job %q failed: %s",
+					jobName,
+					message,
+				)
```
tuunit:
```suggestion
				return fmt.Errorf("credential rotator Job %q failed: %s",	jobName, message)
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+			}
+		}
+
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf(
+				"waiting for credential rotator Job %q: %w",
+				jobName,
+				ctx.Err(),
+			)
```
tuunit:
```suggestion
			return fmt.Errorf("waiting for credential rotator Job %q: %w", jobName, ctx.Err())
```

src/internal/cmd/bootstrap/spoke.go
```diff
@@ -0,0 +1,603 @@
...
+		defer cancel()
+
+		jobLogs := getRotatorJobLogs(logCtx, client, job.Name)
+		if jobLogs != "" {
+			return fmt.Errorf(
+				"%w\ncredential rotator logs:\n%s",
+				err,
+				jobLogs,
+			)
+		}
```
tuunit: why throw an error if you get the logs? why would that be a negative?

tuunit (changes requested):

src/cmd/bootstrap.go
Rana-Ashar: Not quite equivalent when KUBECONFIG is unset assigning os.Getenv directly
would replace the default ~/.kube/config value with an empty string. I kept
the non empty check so the normal CLI default remains the fallback.

fabian-schmitt:
There is a lot of open feedback and it is distracting me while releasing. i will move this back to draft.

Issue #575, opened 2026-08-19
[Feature]: Gateway API


tuunit:
### 📝 Description

We want to support Gateway API in future releases for this some minor code changes and quite a few catalog changes will be necessary.

### 🎯 Use Case

Alternative to Ingress

### 📎 Additional Context

Traefik supports Gateway API:

https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-gateway/

https://github.com/kubara-io/catalogs/tree/main/general/platform-configs/helm/traefik

Furthermore, some code changes will be necessary:

Generalized networking block instead of the `IngressClassName`
https://github.com/kubara-io/kubara/blob/main/src/internal/config/types.go#L62
https://github.com/kubara-io/kubara/blob/main/src/internal/service/types.go#L17-L19

### ✅ Checklist

- [x] I have searched existing issues
- [x] This is not a duplicate

Brilly-Bohyun:
I'll work on this 🫡

Brilly-Bohyun:
Hi @tuunit, I have a local implementation in progress and would like to check that the approach matches your intent before opening PRs.
- Keep Ingress as the default and make Gateway API an opt-in alternative using Traefik.
- Replace the top-level ingressClassName with a networking block containing either ingressClassName or a Gateway reference (name, namespace, optional sectionName), with service-level Gateway overrides.
- Keep Go changes focused on configuration handling, and implement Gateway/HTTPRoute resources and the related authentication, TLS, and DNS settings in the catalogs.

Two points I’d like to clarify:

1. Does “instead of IngressClassName” mean removing the old top-level field and requiring users to update their configuration, or would you prefer an automatic migration?
2. Should the catalogs install Gateway API CRDs during bootstrap, or is documenting them as a prerequisite acceptable for the initial implementation?

Does this scope match what you had in mind?

tuunit:
@Brilly-Bohyun 

I would for now agree with your bullet point list. Of course some of it might change when the PRs are created and I see the full picture.

But overall yes the resources should be handled in the catalogs. Kubara should only provide the necessary configuration.

> 1. Does “instead of IngressClassName” mean removing the old top-level field and requiring users to update their configuration, or would you prefer an automatic migration?

[...]

PR #630, opened 2026-09-18
feat: add Gateway API networking configuration

Brilly-Bohyun:
## 📝 Summary
Add structured cluster routing configuration and service-level Gateway references so catalogs can generate Gateway API routes. Ingress remains the default.

- Support `networking.ingressClassName` or `networking.gateway` (name, namespace, optional sectionName), with complete service Gateway overrides.
- Retain the deprecated top-level `ingressClassName` for existing catalogs. Loading an old Ingress config warns and automatically populates and saves the structured field. Reject conflicting explicit Ingress values.
- Reset Gateway references to the Traefik Ingress profile in local mode.
- Keep controller, route, authentication, TLS, DNS, and CRD resources in the companion catalogs change.

## 🧩 Type of change
- [x] 🔧 CLI / Go code
- [x] 📝 Documentation
- [x] 🧪 Test or CI change
- [ ] ♻️ Refactor / cleanup

## 🧪 Testing
- [ ] CI passed
- [x] Unit tested
  
## 🔗 Additional Context / Related Issues / Tickets
Related to #575 

## ✅ Checklist


docs/content/2_concepts/catalog_templating.md
```diff
@@ -56,6 +56,48 @@ For example:
 - `.cluster.services.traefik.status`
 - `.cluster.services.cert-manager.config.clusterIssuer.name`
 
+#### Networking
+
+`.cluster.networking` contains the current cluster's routing settings.
+
+Templates can read fields such as:
+
+- `.cluster.networking.ingressClassName`
```
tuunit: lets use the opportunity and make this a bit more generic
```suggestion
- `.cluster.networking.ingress.className`
```

this way we can support more ingress specific values from the global config in the future

docs/content/2_concepts/catalog_templating.md
```diff
@@ -56,6 +56,48 @@ For example:
...
+A Gateway reference requires `name` and `namespace`; `sectionName` optionally selects a listener.
+
+For example, this cluster configuration supplies a parent Gateway:
+
+```yaml
+networking:
+  gateway:
+    name: platform
+    namespace: traefik
+    sectionName: websecure
```
tuunit, line 84: I'm wondering if we should keep this implicit way of configuring this or if we should introduce an explicit enum type like:


```suggestion
networking:
  type: <ingress|gateway>
  ingress:
    className: traefik
  gateway:
    name: platform
    namespace: traefik
    sectionName: websecure
```

what do you think?

src/internal/config/defaults.go
```diff
@@ -1,12 +1,52 @@
...
+		networking := cluster.Networking
+		if cluster.IngressClassName != "" {
+			log.Warn().Str("cluster", cluster.Name).Msg("ingressClassName is deprecated; use networking.ingressClassName")
+			if networking.Gateway == nil {
+				if networking.IngressClassName != "" && networking.IngressClassName != cluster.IngressClassName {
+					return fmt.Errorf("cluster %q: ingressClassName conflicts with networking.ingressClassName", cluster.Name)
+				}
+				networking.IngressClassName = cluster.IngressClassName
+			}
+		}
```
tuunit: if we ignore the setting anyways in case of gateway being used/set

wouldn't it be cleaner to just do:

```suggestion
		if cluster.IngressClassName != "" && networking.Gateway == nil {
			log.Warn().Str("cluster", cluster.Name).Msg("ingressClassName is deprecated; use networking.ingressClassName")
			if networking.IngressClassName != "" && networking.IngressClassName != cluster.IngressClassName {
				return fmt.Errorf("cluster %q: ingressClassName conflicts with networking.ingressClassName", cluster.Name)
			}
			networking.IngressClassName = cluster.IngressClassName
		}
```

yes this would only print the warning if ingressClassName is set and used but would make the code less nested and easier to read

src/internal/config/defaults.go
```diff
@@ -1,12 +1,52 @@
...
+		if networking.Gateway != nil && networking.IngressClassName != "" {
+			return fmt.Errorf("cluster %q: networking cannot contain both ingressClassName and gateway", cluster.Name)
+		}
```
tuunit: ah okay with my assumption / suggested change this would likely fail.

I think we should do something differently:

* Remove the support for the surface level field. Meaning the migration should take care of the field contents to be moved to the new struct
* We still need to keep the field for serialization for backwards compatibility for the catalogs but not for the surface level config support
* Therefore introduce an error and kubara failure if someone would manually add the field back to the config


This way we can have one simple check here with a failure at the top if the field has contents. Otherwise we procede with the validation for the new struct and copy the `ingress.className` from the new struct into the cluster struct in-memory

this should make the code cleaner. thinking out loud: we might need to introduce a `kubara catalog migration` command in the future but that is a story for another day

tuunit (changes requested):

Brilly-Bohyun:
@tuunit Thanks for the suggestions. I've updated the configuration to use networking.ingress.className and an explicit networking.type (ingress or gateway).
Migration now moves the legacy ingressClassName into the new structure. 
Reintroducing it alongside networking causes an error.
After validation, the Ingress class is copied into the in-memory compatibility field for existing catalogs, but excluded when saving the config. 

Migration identifies the legacy format by the absence of networking, so removing the entire block makes the config look like legacy input again. 
Does this distinction fit what you had in mind?

docs/content/2_concepts/catalog_templating.md
```diff
@@ -56,6 +56,53 @@ For example:
...
+
+`.cluster.networking` contains the current cluster's routing settings.
+
+Templates can read fields such as:
+
+- `.cluster.networking.type`
+- `.cluster.networking.ingress.className`
+- `.cluster.networking.gateway.name`
+- `.cluster.networking.gateway.namespace`
+- `.cluster.networking.gateway.sectionName`
```
tuunit:
```suggestion
The networking block is defined as follows:

```yaml
networking:
  type: ingress # or gateway
  ingress:
    className: your-class-name # defaults to traefik
  gateway:
    name: your-gateway
    namespace: namespace-of-controller # e.g. traefik
    sectionName: optional-listener-section
```

```

File: src/cmd/generate_test.go [hunk 6]
```diff
@@ -234,11 +234,11 @@ func TestGenerateCmd(t *testing.T) {
...
 			if !tt.wantErr || tt.errContains != "load config" {
 				cluster := config.Cluster{
-					Name:             "test-cluster",
-					Stage:            "dev",
-					IngressClassName: "traefik",
-					Type:             "hub",
-					DNSName:          "test.example.com",
+					Name:       "test-cluster",
+					Stage:      "dev",
+					Networking: &config.ClusterNetworking{Type: config.NetworkingIngress, Ingress: &config.IngressNetworking{ClassName: "traefik"}},
```
tuunit:
```suggestion
					Networking: &config.ClusterNetworking{
						Type: config.NetworkingIngress, 
						Ingress: &config.IngressNetworking{
							ClassName: "traefik",
						},
					},
```

makes it more readable

src/internal/config/migrations/migrations.go
```diff
@@ -48,6 +48,13 @@ func Apply(cwd string, config map[string]any) (bool, error) {
 		migrated = true
 	}
 
+	if config["version"] == ConfigVersionV1Alpha4 {
+		networkingMigrated, err := migrateIngressNetworking(config)
+		if err != nil {
+			return false, fmt.Errorf("migrate Ingress networking: %w", err)
+		}
+		migrated = migrated || networkingMigrated
+	}
```
tuunit: please follow the same convetion as above:
* introduce a isV1Alpha4Config method
* introduce a migrateV1Alpha4Config method
* the migration should return an error or nothing
* in case of isV1Alpha4Config and no error the migration must be true. the logical or seems unncessary to me

src/internal/config/migrations/networking.go
```diff
@@ -0,0 +1,35 @@
+package migrations
+
+import "fmt"
+
+// migrateIngressNetworking upgrades legacy-only configs without changing their version.
+// A networking block marks the structured format; any legacy field beside it is
+// left for config validation to reject instead of being migrated again.
```
tuunit: but this change is a breaking change to the file structure and parsing of the config and therefore requires a version bump of the config file

src/internal/config/defaults_test.go
```diff
@@ -23,18 +24,18 @@ func TestApplyDefaults_ClusterLevelDefaults(t *testing.T) {
...
 			{
-				Name:             "test",
-				Stage:            "production",
-				Type:             "spoke",
-				IngressClassName: "nginx",
-				DNSName:          "test.example.com",
+				Name:       "test",
+				Stage:      "production",
+				Type:       "spoke",
+				Networking: &ClusterNetworking{Type: NetworkingIngress, Ingress: &IngressNetworking{ClassName: "nginx"}},
```
tuunit: please use multiline struct formatting as this makes it hard to read otherwise

src/internal/config/factory.go
```diff
@@ -39,13 +39,13 @@ func NewClusterFromEnvWithCatalog(e *envconfig.EnvMap, catalogOptions catalog.Lo
...
-		SSOOrg:           "<my-org>",
-		SSOTeam:          "<my-team>",
-		IngressClassName: "traefik",
+		Name:       e.ProjectName,
+		Stage:      e.ProjectStage,
+		Type:       "<hub or spoke>",
+		DNSName:    "<subdomain.my-domain.com>",
+		SSOOrg:     "<my-org>",
+		SSOTeam:    "<my-team>",
+		Networking: &ClusterNetworking{Type: NetworkingIngress, Ingress: &IngressNetworking{ClassName: "traefik"}},
```
tuunit: same for this one

src/internal/config/factory_test.go
```diff
@@ -37,13 +37,13 @@ func TestNewClusterFromEnv(t *testing.T) {
...
-		SSOOrg:           "<my-org>",
-		SSOTeam:          "<my-team>",
-		IngressClassName: "traefik",
+		Name:       "kubara-test",
+		Stage:      "dev",
+		Type:       "<hub or spoke>",
+		DNSName:    expectedDNSName,
+		SSOOrg:     "<my-org>",
+		SSOTeam:    "<my-team>",
+		Networking: &ClusterNetworking{Type: NetworkingIngress, Ingress: &IngressNetworking{ClassName: "traefik"}},
```
tuunit: same here

src/internal/config/store.go
```diff
@@ -59,10 +59,17 @@ func (cs *ConfigStore) Load() error {
 		return fmt.Errorf("migration of config failed: %w", err)
 	}
 
+	if err := rejectLegacyIngressConfig(raw); err != nil {
+		return err
+	}
```
tuunit: if we enforce the conversion and version bump as previously mentioned at load time. we can never reach this point and therefore this is deadcode

src/internal/config/store.go
```diff
@@ -59,10 +59,17 @@ func (cs *ConfigStore) Load() error {
...
+	// A fresh value prevents repeated loads from retaining catalog compatibility fields.
+	cs.config = &Config{}
```
tuunit: what does this have todo with this PR? is this even possible that we have repeated loads? every kubara command does one load only, right?

src/internal/config/defaults.go
```diff
@@ -7,6 +7,20 @@ import (
...
+		cluster := &cfg.Clusters[i]
+		if cluster.Networking == nil {
+			cluster.Networking = &ClusterNetworking{}
+		}
+		networking := cluster.Networking
+		if (networking.Type == "" || networking.Type == NetworkingIngress) && networking.Ingress == nil {
+			networking.Ingress = &IngressNetworking{}
+		}
+	}
+}
```
tuunit: shouldn't this be dealt with by applying the defaults?

why is the networking field a pointer?

src/internal/config/store.go
```diff
@@ -73,6 +80,15 @@ func (cs *ConfigStore) Load() error {
 		return fmt.Errorf("decode config: %w", err)
 	}
 
+	// Unknown routing fields must not disappear before schema validation.
+	slices.Sort(metadata.Unused)
+	for _, key := range metadata.Unused {
+		if strings.Contains(key, ".networking.") {
+			return fmt.Errorf("unknown networking field %q", key)
+		}
+	}
```
tuunit: but why? wouldn't this be something we should apply on the decoder level?
if we don't want to decode unknown fields mapstructure already has a feature for that

tuunit (changes requested):

tuunit:
@Brilly-Bohyun thanks for all the work you put in, as this is quite a complex topic I hope you don't get discouraged by a couple of back and forth rounds before we settle on a final solution

docs/content/2_concepts/catalog_templating.md
Brilly-Bohyun: Updated to the YAML example you suggested. I originally listed the template paths, but that didn’t show how the fields fit together in the configuration. The example makes the structure and defaults clearer 👍🏻

Brilly-Bohyun:
Thanks for taking the time to work through this with me, @tuunit. I’m not discouraged—I appreciate the feedback. It helped me see where I focused too much on compatibility and missed the existing migration conventions and readability.

Sorry for missing some of the basics, like breaking up long lines and following the existing formatting conventions. I should have caught those before asking for a review, and I’ll pay closer attention to them going forward. 🫡

